PR #30, opened 2023-10-09
RFC: ideas around dependencies and testing

jettr:
I am uploading this as a means for discussion, especially for the TCG conference coming up.

This RFC gives examples around dependency injection that we have found very helpful in our embedded rust code base.

I hope to also add some more examples around memory arenas before the conference as well.


chrisfenner:
OK this is weird. I did not intend to close this. Taking a look...

jettr:
It is good to close this. It served its purpose as a discussion aid.

PR #58, opened 2024-02-23
Add skeleton session/handle support to client

esmusick:
This starts to add support for sessions and handles to the client. It
* defines some types/consts needed
* defines the session trait and adds basic operations for a password session
* adds routines for using the simplest sessions/handles for commands
This is the bare minimum to support a password session and thus does not support parameter encryption.

I've left using these with a command to a followup just to keep size down.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,6 +12,7 @@ members = [
 
 [workspace.dependencies]
 # Third party dependencies
+arrayvec = { version="0.7.4", default-features=false }
 bitflags = "2.4.2"
 hex-literal = { version = "0.4.1" }
 open-enum = "0.4.1"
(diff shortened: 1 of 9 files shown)


chrisfenner (approved):

Cargo.toml
```diff
@@ -12,6 +12,7 @@ members = [
 
 [workspace.dependencies]
 # Third party dependencies
+arrayvec = { version="0.7.4", default-features=false }
```
oddcoder, line 15: Is there any reason to use arrayvec over tinyvec or smallvec?

My main preference goes to tinyvec since it is free from unsafe code, the only compromise is that we would need to restructure thing such that the type of data we store should implement default.

esmusick: > Is there any reason to use arrayvec over tinyvec or smallvec?

Only personal familiarity :) I'm open to using something else, but implementing default seems a bit awkward: how do you envision doing so? Ideally we would avoid a present-but-unset state that doesn't mean anything other than (probably) a bug somewhere, and I think we do want to keep Session as a trait to allow arbitrary impls, although this might not be essential.

esmusick: Got it, thanks for the discussion! I'm going to leave it as-is for now, since it doesn't seem like a clear improvement to switch, but like you I don't have a preference either way so no objections to changing it in the future if doing so makes sense.

PR #63, opened 2024-03-06
Add CreatePrimary command support.

esmusick:
This serves as an initial use to exercise session/handle support
- Defines requisite structures
- Adds a features-level helper for invoking
- Runs against simulator for testing basic functionality

### DIFF
--- a/client/src/lib.rs
+++ b/client/src/lib.rs
@@ -1,12 +1,12 @@
 #![forbid(unsafe_code)]
 #![cfg_attr(not(test), no_std)]
 use core::mem::size_of;
-use sessions::CmdSessions;
+use sessions::{CmdSessions, PasswordSession};
 use tpm2_rs_base::commands::*;
-use tpm2_rs_base::constants::{TPM2CC, TPM2ST};
+use tpm2_rs_base::constants::{TPM2Handle, TPM2CC, TPM2ST};
 use tpm2_rs_base::errors::{TpmError, TpmResult, TssTcsError};
 use tpm2_rs_base::marshal::{Marshal, Marshalable, UnmarshalBuf};
-use tpm2_rs_base::{TpmiStCommandTag, TpmsAuthResponse};
+use tpm2_rs_base::{TpmiRhHierarchy, TpmiStCommandTag, TpmsAuthResponse};
 
 pub const CMD_BUFFER_SIZE: usize = 4096;
 pub const RESP_BUFFER_SIZE: usize = 4096;
@@ -17,11 +17,24 @@ pub trait Tpm {
     fn transact(&mut self, command: &[u8], response: &mut [u8]) -> TpmResult<()>;
 }
 
+// These methods are syntatic sugar for calling run_command* with a specific command.
+
 pub fn get_capability<T: Tpm>(
     tpm: &mut T,
     command: &GetCapabilityCmd,
 ) -> TpmResult<GetCapabilityResp> {
     run_command(command, tpm)
+}
+
+pub fn create_primary<T: Tpm>(
+    tpm: &mut T,
+    command: &CreatePrimaryCmd,
+    primary_handle: TpmiRhHierarchy,
+) -> TpmResult<(CreatePrimaryResp, TPM2Handle)> {
+    let mut session = PasswordSession::default();
+    let mut sessions = CmdSessions::default();
+    sessions.push(&mut session);
+    run_command_with_handles(command, primary_handle, sessions, tpm)
 }
 
 #[repr(C)]
--- a/base/src/commands.rs
+++ b/base/src/commands.rs
@@ -1,5 +1,8 @@
 use crate::constants::{TPM2Cap, TPM2CC, TPM2PT, TPM2SU};
-use crate::{Marshal, Marshalable, UnmarshalBuf};
+use crate::{
+    Marshal, Marshalable, TPM2Handle, Tpm2bCreationData, Tpm2bData, Tpm2bDigest, Tpm2bName,
+    Tpm2bPublic, Tpm2bSensitiveCreate, TpmiRhHierarchy, TpmtTkCreation, UnmarshalBuf,
+};
 use crate::{TpmiYesNo, TpmlDigest, TpmlPcrSelection, TpmsCapabilityData};
 
 /// Trait for a TPM command transaction.
@@ -65,4 +68,29 @@ pub struct PcrReadResp {
     pcr_update_counter: u32,
     pcr_selection_out: TpmlPcrSelection,
     pcr_values: TpmlDigest,
+}
+
+#[repr(C)]
+#[derive(Clone, Copy, PartialEq, Debug, Marshal)]
+pub struct CreatePrimaryCmd {
+    pub in_sensitive: Tpm2bSensitiveCreate,
+    pub in_public: Tpm2bPublic,
+    pub outside_info: Tpm2bData,
+    pub creation_pcr: TpmlPcrSelection,
+}
+impl TpmCommand for CreatePrimaryCmd {
+    const CMD_CODE: TPM2CC = TPM2CC::CreatePrimary;
+    type Handles = TpmiRhHierarchy;
+    type RespT = CreatePrimaryResp;
+    type RespHandles = TPM2Handle;
+}
+
+#[repr(C)]
+#[derive(Clone, Copy, PartialEq, Debug, Marshal)]
+pub struct CreatePrimaryResp {
+    pub out_public: Tpm2bPublic,
+    pub creation_data: Tpm2bCreationData,
+    pub creation_hash: Tpm2bDigest,
+    pub creation_ticket: TpmtTkCreation,
+    pub name: Tpm2bName,
 }
--- a/base/src/lib.rs
+++ b/base/src/lib.rs
@@ -294,6 +294,18 @@ impl TryFrom<u32> for TpmiRhNvIndex {
     }
 }
 
+/// TpmiRhHierarchy represents a hierarchy selector handle (TPMI_RH_HIERARCHY).
+/// See definition in Part 2: Structures, section 9.13.
+#[open_enum]
+#[repr(u32)]
+#[rustfmt::skip] #[derive(Debug)] // Keep debug derivation separate for open_enum override.
+#[derive(Copy, Clone, PartialEq, Default, Marshal)]
+pub enum TpmiRhHierarchy {
+    Owner = TPM2Handle::RHOwner.0,
+    Platform = TPM2Handle::RHPlatform.0,
+    Endorsement = TPM2Handle::RHEndorsement.0,
+}
+
 /// TpmiShAuthSessions represents handles referring to an authorization session (TPMI_SH_AUTH_SESSION).
 /// See definition in Part 2: Structures, section 9.8.
 #[repr(transparent)]
@@ -322,6 +334,11 @@ impl TpmiShAuthSession {
 #[repr(transparent)]
 #[derive(Clone, Copy, PartialEq, Debug, Default, Marshal)]
 pub struct TpmiEccCurve(TPM2ECCCurve);
+impl TpmiEccCurve {
+    pub fn new(curve: TPM2ECCCurve) -> TpmiEccCurve {
+        TpmiEccCurve(curve)
+    }
+}
 
 /// TpmiYesNo is used in place of a boolean.
 /// See TPMI_YES_NO definition in Part 2: Structures, section 9.2.
@@ -575,6 +592,14 @@ impl TpmuName {
     pub const fn union_size() -> usize {
         size_of::<TpmuName>() - align_of::<TpmuName>() - size_of::<u16>()
     }
+}
+
+#[repr(C)]
+#[derive(Clone, Copy, PartialEq, Debug, Marshal)]
+pub struct TpmtTkCreation {
+    tag: TPM2ST,
+    hierarchy: TpmiRhHierarchy,
+    digest: Tpm2bDigest,
 }
 
 #[repr(C)]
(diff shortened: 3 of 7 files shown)


oddcoder:
I have a question, related to your pr but not directly!

What is the entry point for an arbitrary TPM command?

I see `TPMClient` wraps `Tpm` offers high level rust API, It first construct a command, and send it to custom handler (in this case `create_primary`, which calls the other `create_primary` which calls `run_command_with_handles`.

But `run_command_with_handles` basically marshals the data and send it to `Tpm::transact`. Which is either not implemented yet or not part of this project.


What confuses me is 
A) I cannot find the business logic of `create_primary`?
I was under the impression that `Tpm::transact` should have been the entry point, where it would first unmarshal the command, manage the internal state, and then have requirements (in form of trait) about the chip capabilities (hashes + aes ...etc). What Am I missing?

a-pronin:
> I was under the impression that `Tpm::transact` should have been the entry point, where it would first unmarshal the command, manage the internal state, and then have requirements (in form of trait) about the chip capabilities (hashes + aes ...etc). What Am I missing?

`Tpm::transact` is the client-side client-specific method that knows how to send the command blob to the TPM (e.g. talk to /dev/tpmXXX device; but can be something much more involved like sending over the network to a different device that actually has TPM if that's the use case this client is written for; but in any case the marshaling logic shouldn't depend on the specific of the transport layer). The unmarshalling etc is done on the other side of this connection.

esmusick:
Just to add a couple of comments to Andrey's excellent explanation:
- We do have an implementation of Tpm::transact but it's only for talking to the TPM simulator ([here](https://github.com/tpm-rs/tpm-rs/blob/dda85e59712cde22009a1eacc56dfe966e63f36b/client/tests/common/mod.rs#L111)). Real use cases will need to supply their own to use the client.
- If you're looking for the service-side entry point, I think [this](https://github.com/tpm-rs/tpm-rs/blob/dda85e59712cde22009a1eacc56dfe966e63f36b/service/src/service.rs#L64) is probably the closest thing, although it's mostly a skeleton.

client/src/lib.rs
```diff
@@ -17,13 +17,26 @@ pub trait Tpm {
...
     run_command(command, tpm)
 }
 
+pub fn create_primary<T: Tpm>(
+    tpm: &mut T,
+    command: &CreatePrimaryCmd,
+    primary_handle: TpmiRhHierarchy,
+) -> TpmResult<(CreatePrimaryResp, TPM2Handle)> {
+    let mut session = PasswordSession::default();
+    let mut sessions = CmdSessions::default();
```
oddcoder, line 35: I am not 100% sure how sessions work in tpms yet, but this does not feel right to me, the three session handles are propagated down from user to ESAPI as here https://github.com/tpm2-software/tpm2-tss/blob/ba2ea649212388f6d928f6598a6f388023a77599/src/tss2-esys/api/Esys_CreatePrimary.c#L96C1-L97C22.

and I imagine it will be the same down to tpm ? am I missing something?

esmusick: Yeah so this is currently assuming you want a single password session for simplicity. When/if there's a use case for more sessions, other helpers can be added or more parameters can be plumbed through this one.

oddcoder: I am not fully convinced what use case you have in mind, but aside from this small thing, I think that patch look fine to me,I built few things on top and testing.


@chrisfenner  will it be possible to approve this patch soon as I already have a patch that builds on top.

chrisfenner:
Sorry for the languishing here. Let me add @trent-reed to take a look (his grasp of Rust far surpasses mine)

trent-reed:
Yeah, sorry for the delay here - there's been a transfer of responsibilities from Emily to me for this project, and I've just been reading up on what I need to read up on TPMs to start contributing (I know rust, I work in security, but I'm fairly new to TPMs). Anyways, I probably should have sent an update email earlier, even if it was just "please give me time".

So I can say this, I'm going to be proposing a lot of major restructuring for this project in July. I'm working on some PRs to clean things up and fix some issues, but some of the changes required to clean things up are quite massive (and you know how it goes, one thing leads to another when you start pulling threads). So I'd prefer if we didn't take too many changes until after that clean-up work happens.

There's some structures of the code that I'm not too convinced by, and I want to deal with that before the project grows much bigger than it already is.

Anyways, I'm fine taking the work that has already been done here, and reviewing whatever you may have that builds on top of this @oddcoder, but let's not go much further than that until other cleanup is done. I'll send out some PRs and ask for some help on that cleanup in the coming month.

Again, my apologies for the delay; I wanted to come in here with a better idea of the long-term direction of the project before I started opening issues on improvements, but probably I should have communicated that earlier.

oddcoder:
will it be possible to open an issue for those modifications, especially if they are big ?  Or at least is there a work-group for discussions regarding tpm-rs design ?

trent-reed:
Yes I plan on doing a write-up with my suggestions for these changes and then filing issues for them. There will be some discussion on these changes though, I don't want to just push a bunch of PRs without talking about them first, but I think the current design is easily maintainable and there's some broken windows I want to address before we start adding a bunch of new commands.

Please give me a little more time to finalize some thoughts on my end for some of the remaining things that I'm investigating, and then I'll start being more active in this repo and making some more posts regarding the design and the future of this project.

trent-reed:
Just in case anyone is reading over email only; I edited my original post above. It erroneously said "I think the current design is easily maintainable", but I meant to say "I **don't** think the current design is easily maintainable". :P

Anyways, I want to focus on clean-up and fixing some of these issues first, otherwise it'll be hard to fix them later after we've implemented a ton of commands and tests. I'll start a discussion about these issues and file bugs as soon as I am able to.

client/src/lib.rs
trent-reed: Frankly, I agree that this seems incorrect. I'll make some suggestions in the near future for how to abstract this client.

trent-reed:
Just a note @oddcoder - I started a discussion about the changes that I'd like to make for the project moving forward. Feel free to respond there if you have thoughts! I think my current stance is that I don't want to take this PR, as we'll probably be refactoring the client code quite a lot.

oddcoder:
@esmusick If I understand correctly, you are no longer working on this patch. If that is indeed the case, then I would be always rebasing it against main until it is right time to merge, while doing tweaks here and there as necessary.

If there is any objection to that, or any misunderstanding on my side please let me know.

trent-reed:
My 2 cents is that we should drop this patch. There's going to be some big changes to the way the client object works, and I think it'll be incompatible with what we have today. I'd really rather not take any more commands until we have the fundamentals of this repo better formed.

esmusick:
I'm happy to jump on and do basic rebasing, etc. but I'm afraid I don't have bandwidth right now for more involved development. @oddcoder, please feel free to take this PR over or copy whatever is useful into your own!

oddcoder:
> My 2 cents is that we should drop this patch. There's going to be some big changes to the way the client object works, and I think it'll be incompatible with what we have today. I'd really rather not take any more commands until we have the fundamentals of this repo better formed.

Under normal circumstances, people contributing code cannot hope to expect locking the whole project for unspecified period of time, until their ideas turn to fully working patch-set. Normally, it will be the contributor's responsibility to keep up with the  moving project until their work is ready for submission.

I tentatively accepted having to do it the other way around because there are not many active contributes at the present time, and I don't want to make anyone's life harder. Had there been more people, some may have disagreed to the approach and it would be 100% within their right.

However, it is not reasonable to not only lock the project, but also halt all future developments until one contributor thinks it is ready. Not only that this is inefficient use of everyone's time, but is discourages other people from joining in.

I understand you are quite dissatisfied with this patch, so can you please point out any problematic part, so that we can fix it? This would be preferable to offering alternative implementation at a later, undefined time, which may prolong inadvertently.

I'm all for collaboration and hope that we can find a balanced approach to allow everyone to make progress on this community project.

trent-reed:
Yeah, that's fair - I think my fear is we still have open questions about some of the fundamental base types (like what to do about validation, error types still aren't complete, some of the abstractions aren't fully-formed for the more complex base types, places where we're depending on private functionality that would not be implementable by someone outside of the main `lib.rs` module, unclear testing abstractions and standards, etc), so it feels risky and like we will be producing a lot of unnecessary additional technical debt to start rushing towards implementing a bunch of commands.

I do agree that I've been heavy-handed in my approach to clean things up, though. I'm sure I've stepped on some toes here, so I apologize for that. It feels to me like there's a lot of foundational stuff that either isn't designed, or is growing naturally and doesn't feel like it's growing the right direction to me.

As for this PR, putting aside all of the other concerns I've raised for the repo, I'd say this:

1. I think we should not have a new functions for every command - the TPM commands are very much set up like RPCs which look like `Request`/`Response` objects, so there should be no reason why we can't just have one `run_command()` function or whatever you want to call it.
2. Sessions aren't currently properly handled, I think we need to be able to mark-up the required sessions on the command request (or some other struct type) and pass them through there. I think we're going to create a lot of trouble if we do a bunch of `Session::default()` everywhere instead of properly plumbing it through.
3. I think the core TPM type can be fully-defined (as in not a trait), and it takes a "connection" type should be a trait to allow for different backends to communicate with a TPM, that way clients can provide their own connection type trivially without every instance of a TPM being a new instance of the code.

Where exactly to separate the required sessions from the optional ones isn't the most clear to me, but apparently some handles have markup in the documentation (`@`) when they *require* an authorization session for the handle, so we can probably separate in that way (this segmentation proposed to me by @chrisfenner, and it is what the go tpm2 code does, see this for example: https://github.com/google/go-tpm/blob/main/tpm2/test/sealing_test.go)

The ideal (IMO) TPM client object looks something like this in my mind:

```rs
trait TpmConnection {
  // Not sure the exact result type, but probably not TpmResult.
  // Definitely some TSS result, but TPM is not right here.
  // Not totally sure if raw buffers is the way to go, but
  // we can punt on this for the time being.
  fn transact(&mut self, command: &[u8], response: &mut [u8]) -> TssResult<()>;
}

pub struct TpmClient<C> {
  connection: C,
}

impl<C> TpmClient<C>
where C: TpmConnection {
  pub fn new(connection: C) -> Self {
    Self { connection }
  }
  pub fn send<R: TpmCommandRequest>(request: R) -> TssResult<R::Response>;
}
```

---

Then, a real implementation can:

```
let connection = TpmTcpConnection::new(...)?;
let client = TpmClient::new(connection);

let request = SomeCommandRequest {
  handles: ...,
  params: ...,
  sessions: ...,
};
let response = client.send(&request)?;
```

Then for testing you can have one-of:

```
let mock_connection = MockTpmConnection::new();
let client = TpmClient::new(connection);
// Maybe a convenience alias of `MockTpmClient` also exists.
// Use the mock...
```

or perhaps in the future we can provide a reasonable simulator-like fake for more complex code:

```
let fake_connection = FakeTpmConnection::new();
let client = TpmClient::new(connection);
// Maybe a convenience alias of `FakeTpmClient` also exists.
// Use the mock...
```

---

I'll stop being so heavy-handed, if you want to take this PR we can take it. Sorry for the trouble.

oddcoder:
Thanks a lot for the feedback!

> I think we should not have a new functions for every command - the TPM commands are very much set up like RPCs which look like Request/Response objects, so there should be no reason why we can't just have one run_command() function or whatever you want to call it.

There are some merits to having TPM commands as individual functions, that is having rusty types like `bool` instead of the TPM ones like `TPMYESNO`. Our internal plans was for the combination of client/client-features to serve as lowest layer  of TSS (functionality wise but not API wise).

> I think we need to be able to mark-up the required sessions on the command request (or some other struct type) and pass them through there. I worry that we will create a lot of trouble if we do a bunch of Session::default() everywhere instead of properly plumbing sessions through.

I totally agree with that. But I didn't want to push for this, and thought I'd wait till I have proper use-case, but if you agree on that, we are aligned.

> I think the core TPM type can be fully-defined (as in not a trait), and it takes a "connection" type should be a trait to allow for different backends to communicate with a TPM.

I am not sure if that offers real advantage. I might be missing something though.


> We can pass the required sessions and if you fail to present a required session it should be a compiler error somehow (I believe the Rust type system is comprehensive enough to allow this). Additional optional sessions may be provided as well it's not clear how type-safe we can make this but the session limit is so if we can it'd be nice to make providing more sessions than that a compiler error (this I'm less sure is possible, especially in cases where there is an optional handle).

I think both are doable. The only concern is that we might end up with many different types for various configurations of sessions. If you have no objection (and if you are not really working on that now) I can prototype something this week.

oddcoder:
> but I'm afraid I don't have bandwidth right now for more involved development. @oddcoder, please feel free to take this PR over or copy whatever is useful into your own!

Thanks for the heads up, and I wish you best of luck in your future endeavors.

Closing in favor of https://github.com/tpm-rs/tpm-rs/pull/78

trent-reed:
> There are some merits to having TPM commands as individual functions, that is having rusty types like `bool` instead of the TPM ones like `TPMYESNO`. Our internal plans was for the combination of client/client-features to serve as lowest layer of TSS (functionality wise but not API wise).

We should talk about this, I don't think I'm on the same page here. We can continue this discussion in the new PR though.

> I am not sure if that offers real advantage. I might be missing something though.

There are some advantages, but it depends on what you prioritize. We can discuss the pros and cons in the new PR. Thanks!

> I think both are doable. The only concern is that we might end up with many different types for various configurations of sessions. If you have no objection (and if you are not really working on that now) I can prototype something this week.

This is what I wanted to spend time working on before I commented more on the direction, but I think I've seen enough to at least say we can do something like this, just that I don't know exactly what it'll look like. If you have time to look into it and propose a solution, I'd be happy to take a look and review it!

chrisfenner:
From my experience with https://github.com/google/go-tpm, the TPM interface lends itself much better to an RPC-like request/response structure pair design, vs turning each command into a function. The parameters are both numerous and highly nested. We found the hard way that if you try to turn each TPM command into a function, you will end up making choices for your users. Then later you'll find some users who need different choices, and you have to add overloads on your interface.

Issue #65, opened 2024-03-21
Code organization idea: rename "base" crate to "tpm2" and use {tpm2b, tpmt, tpms, tpmi, tpma} submodules


josephlr:
I used to maintain prototype Rust TPM2 bindings at https://github.com/josephlr/tpm-rs and published this prototype as the [`tpm2` crate on crates.io](https://crates.io/crates/tpm2). I haven't had a lot of time to work on that crate recently, and @chrisfenner told me about this actively-maintained crate. I think the best course of action would be for this project to publish as the `tpm2` crate instead. I can transfer/add people on crates.io if necessary. I'm also happy to help some with this project (although I don't work on TPM2 stuff at work currently).

If folks think publishing as `tpm2` makes sense, this opens up a lot of ergonomics improvements for the naming of types. This builds on the ideas in #15 to make things more Rust-like and easier to read. For example:
  - `tpm2_rs_base::TpmuPublicId` becomes `tpm2::tpmu::PublicId`
  - `tpm2_rs_base::TpmaNv` becomes `tpm2::tpma::Nv`
  - `tpm2_rs_base::constants::TPM2_SHA256_DIGEST_SIZE` becomes `tpm2::SHA256_DIGEST_SIZE`

I moved some of my old prototype code to the new layout, and I think it makes everything more ergonomic. Expand the sections below for examples:

<details><summary>Crate Root (tpm2)</summary>
These would be values accessible at the crate root. This would include "core" type definitions, constants, and aliases.

```rust
pub const SHA1_DIGEST_SIZE: u32 = 20;
pub const SHA256_DIGEST_SIZE: u32 = 32;
pub const SHA384_DIGEST_SIZE: u32 = 48;
pub const SHA512_DIGEST_SIZE: u32 = 64;
pub const SM3_256_DIGEST_SIZE: u32 = 32;

/// TPM_ALG_ID
#[derive(Clone, Copy, PartialEq, Eq, Debug)]
#[repr(u16)]
pub enum Alg {
    Rsa = 0x0001,
    Tdes = 0x0003,
    Sha1 = 0x0004,
    Hmac = 0x0005,
    Aes = 0x0006,
    Mfg1 = 0x0007,
    KeyedHash = 0x0008,
    Xor = 0x000A,
    Sha256 = 0x000B,
    Sha384 = 0x000C,
    Sha512 = 0x000D,
    Sha256_192 = 0x000E,
    Null = 0x0010,
[...]

PR #74, opened 2024-07-30
Flatten the crate roots so that there are no nested crates.

trent-reed:
This change also names the directories the same thing as the crate. In a later PR we will add READMEs to the crate directories to better document them.

For justification, see: https://github.com/orgs/tpm-rs/discussions/18

Some additional changes:

* I renamed tpm2-rs-features-client to tpm2-rs-client-features so that it composes nicer with the tpm2-rs-client crate.
* There were some comments suggesting that we were hiding crates like unionify-derive, but you can't "hide" public crates in rust. So I removed those comments and exposed the crate (that's what'll happen anyways on publish).
* Ordered the dependencies lists so that they are sorted.
* Had to update a few using statements to account for the crate name changes.
* Updated some crates from having paths to their deps to using the workspace setting.


trent-reed:

Cargo.toml
```diff
@@ -1,14 +1,15 @@
...
 resolver = "2"
 members = [
-    "base",
-    "client",
-    "client/feature",
-    "errors",
-    "marshal",
-    "marshal/derive",
-    "service",
-    "unionify"
```
oddcoder, line 11: Prefixing every directory with `tpm2_rs` feel unnecessary. I strongly prefer prefix-free directories. This way we won't won't have to `cd tpm2-rs-` every time we want to go to somewhere. It makes paths longer than they need to be, especially that we are in tpm-rs repository anyways.

trent-reed: I find the clarity of using the actual crate name to be a nice property. If other directories end up in our root like `docs` or `norms` or something like that, it'll be clear what's what, and it keeps them organized closely together.

There's one other reason I think we might want to prefix these directories: A future exists where we make the base crate suffix-less (see: https://github.com/tpm-rs/tpm-rs/issues/65).

I think in that case the list of crates in this directory will be:

*   tpm2
*   tpm2-client
*   tpm2-client-features
*   tpm2-errors
*   tpm2-marshal
*   tpm2-marshal-derive
*   tpm2-service
*   tpm2-unionify
*   tpm2-unionify-derive

trent-reed: If you want, we can do prefix-free for now but revisit this after https://github.com/tpm-rs/tpm-rs/issues/65. I still prefer having the prefix, but if you really don't like it I won't block on it.

oddcoder: I personally don't prefer it. Think about it, it adds no extra information, even docs is technically `tpm2-rs-docs`. So what `tpm2-rs` would signify that this is code while other parts are not code. However I don't think this distinction is really necessary at this point, and people can just `cd` in and see what is in there.

trent-reed: docs wouldn't be `tpm2-rs-docs`, it would have just been `docs`, like this:

* docs  // Because this is not a crate
* tpm2-rs-base
* tpm2-rs-client
* tpm2-rs-client-features
* tpm2-rs-... etc. etc.

I disagree, but not strongly enough to push back about it. I'll update the PR.

oddcoder (approved):
I gave it a quick look again and everything looks good to me. Thanks a lot!

PR #76, opened 2024-08-01
Break up the monolithic modules into submodules

trent-reed:
The current structure of code makes it difficult to maintain, and also makes it hard to see what's where and what manually-implemented traits or impls the types have on them.

This PR does not aim to fix all of the crates in one go, instead it will focus on the most egregious crate, `tpm2-rs-base`. However, because there are a number of issues, we did have to make more changes than I would have liked - but it would have been difficult and time-consuming to split these into separate PRs. I am hopeful that this kind of PR will be a rare occurrence after we get some cleanup done on the repo.

See: https://github.com/orgs/tpm-rs/discussions/19

Changes:

* Each command has been moved into its own submodule so that the types relating to the command are easily organized (I'm open to discuss if this is too much separation; but will have requests if we put them all back in the same `commands.rs` file).
* The base types have been moved into modules separated by their common prefix.
* Helper types/functions/macros closely associated with a specific prefix have also moved into the related prefix module.
* Most tests have been moved into separate files named `tests.rs`; this is easier to read and maintain, but the tests are still monoliths that need moved closer to their tested code. This is work for a later PR.
* The types have been organized alphabetically based on their typename within the submodules.
* I've stopped publicly exporting other crates from the base crate - I'm okay with cross-crate re-exports for derive macros, but for anything else it feels very gross to me.
* Corrected the dependencies to account for the new source for these no longer re-exported crates.
* Made the enum marshalable functions an actual thing instead of requiring that types reach into each-other's private functions to access functionality like `discriminant`, `try_marshal_variant`, and `try_unmarshal_variant`.
* The `Marshal` derive had accidentally required users export things in their own crate in order to build. I fixed this where I could by using fully-qualified path names, but the more correct fix requires some more major refactoring that I don't want to do in this same PR.
* `Marshalable` now just has a supertrait of `Sized`; I can't imagine a scenario where we don't want this.
* Got rid of the `Command` type alias for command codes and canonicalized on `TPM2CC` since that name is closer to what the specification calls it (though I don't like the format of this typename... I'll be making a suggestion on proper naming in a later PR).


oddcoder:
Hello Trent,

This PR is huge with lots of shuffling and  some logic modification, There is even changes to some `cargo.toml` files. I would suggest spitting it into separate smaller pull requests, that would make it easier for initial review, later reviews, and for anyone who ever have to bisect to catch accidental bugs. I imagine it can be impractical to do it per structure, so would per file (or group of similar types that share the same theme) be reasonable?

I wasn't able to review the code in details I have few general comments:
- There are lots of decorative comments similar to:

```rust
// =============================================================================
// USES
// =============================================================================

use crate::types::TPM2CC;
use tpm2_rs_marshal::Marshalable;
```

They feel unnecessary.

- There are places where you describe a module using files `foo.rs` and `foo/*.rs` and places where you do `foo/{*|mod}.rs`
  It will be nice only use `foo/{*|mod}.rs` for no reason other than uniformity and that this is what the whole project is using. It is just easier to always go for foo/mod.rs when I want the root module file instead of trying both ways. 
 
 - After splitting, please add the commit description to the commit itself, this way the description stays in the git log.

 I am truly sorry if those requests sound pedantic, but it just makes reviewing easier for me and hopefully for everyone else who might review or even audit the code in the future. I will shortly upload a copy of style guidelines to align everyone's expectations (including myself).

trent-reed:
> This PR is huge with lots of shuffling and some logic modification, There is even changes to some `cargo.toml` files. I would suggest spitting it into separate smaller pull requests, that would make it easier for initial review, later reviews, and for anyone who ever have to bisect to catch accidental bugs. I imagine it can be impractical to do it per structure, so would per file (or group of similar types that share the same theme) be reasonable?

Yup, actually I feel a bit bad about this PR - we haven't worked together too much yet, so you'll have to trust me when I say that I usually do produce more targeted, smaller PRs. Anyways, splitting the code is difficult because it means needing a lot of other changes too. So before I split up the PR I wanted to see if I could just review it as-is, but I definitely agree this is hard to review. Sorry about this, I'll split it up.


> I wasn't able to review the code in details I have few general comments:
> 
> * There are lots of decorative comments similar to:
> 
> ```rust
> // =============================================================================
> // USES
> // =============================================================================
> 
> use crate::types::TPM2CC;
> use tpm2_rs_marshal::Marshalable;
> ```
> 
> They feel unnecessary.

Ah, yeah - I'm a very visual person, so having clear sections for the code makes it easy for me to see where things are and to keep them organized in the future, but I will admit that's a bit of a personal quirk and that not everyone seems to need that - I'll remove them, but I may ask for decorative comments in certain cases in the future where we have lot's of code and visual comment separators are more impactful. For here, I agree that they're not necessary.

> * There are places where you describe a module using files `foo.rs` and `foo/*.rs` and places where you do `foo/{*|mod}.rs`
>   It will be nice only use `foo/{*|mod}.rs` for no reason other than uniformity and that this is what the whole project is using. It is just easier to always go for foo/mod.rs when I want the root module file instead of trying both ways.

Actually, I did not add any `mod.rs`, but I had to clean up existing `mod.rs` files.

I don't want to only use `mod.rs` files - that gets messy fast. We haven't talked about proper module naming format yet, so maybe we need to. How about this:

*   If a module is a leaf module (e.g. no submodules), then it should be defined as `<name>.rs`.
*   If a module is a stem module (e.g. has submodules), then it should be defined as `<name>/mod.rs`.

It's not totally clear if that's what you mean by the above or now, but I want to be really clear on the expectations here - does that sound reasonable?

> * After splitting, please add the commit description to the commit itself, this way the description stays in the git log.

I was going to do that after the squash-rebase-push operation, since I figured there'd be discussion.

> I am truly sorry if those requests sound pedantic, but it just makes reviewing easier for me and hopefully for everyone else who might review or even audit the code in the future. I will shortly upload a copy of style guidelines to align everyone's expectations (including myself).

I don't think this is pedantic, it's appropriate - I'd like to review the code styleguide though, to make sure that we agree.

oddcoder:
>    - If a module is a leaf module (e.g. no submodules), then it should be defined as <name>.rs.
>    - If a module is a stem module (e.g. has submodules), then it should be defined as <name>/mod.rs.

I totally agree with you on that, and that is exactly what I meant, but that is not the state of the current code.
For instance, notice the file [base/src/types.rs](https://github.com/tpm-rs/tpm-rs/pull/76/files#diff-7ee0b3455c12af9b6f4b68a8f0e876b39922574202289761073f4c6d316259e9) is not a leaf module, it has submodules of its own like [base/src/types/tpm.rs](https://github.com/tpm-rs/tpm-rs/pull/76/files#diff-b600f14a1b58e9bacc49dad6b89fdae33cd4541b8d5bcb27d2ef6779e0f0f4d7) among others. I prefer if we standardize the usage of   `base/src/types/mod.rs` instead of `base/src/types.rs` in this particular case.


> I was going to do that after the squash-rebase-push operation, since I figured there'd be discussion.

I am not sure if it is good idea to squash this PR, there are lots of independent moving knobs here. I prefer if we at least split it into smaller commits, and not squash them but preserve the history. I took the liberty of splitting (and reviewing) the parts that touch `marshal`, `marshal-derive` into 4 patches in  local branch: https://github.com/oddcoder/tpm-rs/commit/8bc4ec77c374d70305666569726bd9abf969c676, https://github.com/oddcoder/tpm-rs/commit/42f7254b1f68bb1777bcc7d5eb571c23323f66b0, https://github.com/oddcoder/tpm-rs/commit/47d627bf5d0b3ed42ff672604fda68a668ca825a, and https://github.com/oddcoder/tpm-rs/commit/b132f6118a51b874a8bb167ec0d455dd39cc220c. I paid extra care to not do any modification you didn't make, and instead suggest them here. And I'd argue, that each of those commits will take less than 5 minutes to review, at least that is how I feel about smaller commits.
I hope you can adopt them (and do as many modifications to them as you see fit) and split the rest of this PR in similar fashion.

Regarding the first three, I think they are fine. Maybe for the second one, it is more of common rust pattern to put tests separate `tests` folder outside `src`, and I think it should work in this case, so feel free to do that if you agree with the idea.

Regarding the 4th patch. I Agree that implementing `discriminant`, `try_marshal_variant`, and `try_unmarshal_variant` as methods is not ideal. But Implementing them at a public trait makes them publicly accessible to users (those who derive `Marhallable`). Those functions should be treated as implementation details and should not be made into public API.

I think the correct approach would be using anonymous modules that contains  `discriminant`, `try_marshal_variant`, and `try_unmarshal_variant` as functions, and contains the `Marshalable` trait implementation. Something along the lines of:

```rust
const _: () = {
    impl tpm2_rs_marshal::Marshalable for #name  { /*stuff goes here */}
    fn discriminant(&self) -> #prim { /* stuff */}
    fn try_marshal_variant(&self, buffer: &mut [u8]) -> tpm2_rs_marshal::exports::errors::TpmRcResult<usize> {/*stuff*/}
    fn try_unmarshal_variant(selector: #prim, buffer: &mut tpm2_rs_marshal::UnmarshalBuf) -> tpm2_rs_marshal::exports::errors::TpmRcResult<Self> { /*stuff*/}
};
```

It might happen that `rustc` complains about the anonymous modules, I don't remember why did it usually do that, but if it (still) does for this particular case, we can also us named modules. so long as the variant-related logic stays private.

Let m know what you think.

trent-reed:
> I totally agree with you on that, and that is exactly what I meant, but that is not the state of the current code. For instance, notice the file [base/src/types.rs](https://github.com/tpm-rs/tpm-rs/pull/76/files#diff-7ee0b3455c12af9b6f4b68a8f0e876b39922574202289761073f4c6d316259e9) is not a leaf module, it has submodules of its own like [base/src/types/tpm.rs](https://github.com/tpm-rs/tpm-rs/pull/76/files#diff-b600f14a1b58e9bacc49dad6b89fdae33cd4541b8d5bcb27d2ef6779e0f0f4d7) among others. I prefer if we standardize the usage of `base/src/types/mod.rs` instead of `base/src/types.rs` in this particular case.

Yeah, there wasn't a set style in this code, and I do know of projects who just prefer to use the "new" structure of `<module-name>.rs` with submodules in `<module-name>/...`; so I wasn't sure which y'all would have preferred so I went with the "new" module style instead.

I think having the codestyle guide will help us avoid this kind of churn in the future, so thank you for putting that together.

I will say there will be exceptions to the rule I stated above, most notably with integration tests. Integration tests very confusingly are registered per-rs-file in the tests directory, so they will need to remain `tests/<test-name>.rs` with submodules that are `tests/<test-name>/...`; so please be aware of that special case.

More info: https://doc.rust-lang.org/rust-by-example/testing/integration_testing.html

> I am not sure if it is good idea to squash this PR, there are lots of independent moving knobs here. I prefer if we at least split it into smaller commits, and not squash them but preserve the history. I took the liberty of splitting (and reviewing) the parts that touch `marshal`, `marshal-derive` into 4 patches in local branch: [oddcoder@8bc4ec7](https://github.com/oddcoder/tpm-rs/commit/8bc4ec77c374d70305666569726bd9abf969c676), [oddcoder@42f7254](https://github.com/oddcoder/tpm-rs/commit/42f7254b1f68bb1777bcc7d5eb571c23323f66b0), [oddcoder@47d627b](https://github.com/oddcoder/tpm-rs/commit/47d627bf5d0b3ed42ff672604fda68a668ca825a), and [oddcoder@b132f61](https://github.com/oddcoder/tpm-rs/commit/b132f6118a51b874a8bb167ec0d455dd39cc220c). I paid extra care to not do any modification you didn't make, and instead suggest them here. And I'd argue, that each of those commits will take less than 5 minutes to review, at least that is how I feel about smaller commits. I hope you can adopt them (and do as many modifications to them as you see fit) and split the rest of this PR in similar fashion.

To be clear, I meant that *if* this PR were to be taken, I would have squashed and rebased with a final message. I did say that I would break this into other PRs, so I plan on doing that.

I only left this open for the conversation. But to avoid more confusion I'll just close this PR now.

> Regarding the first three, I think they are fine. Maybe for the second one, it is more of common rust pattern to put tests separate `tests` folder outside `src`, and I think it should work in this case, so feel free to do that if you agree with the idea.

The `tests` directory outside of `src` is for integration tests, not unit tests (and the two different tests do actually have different properties of how they're run). I think it's a fairly common practice for unit tests to be in `src`, and then whether or not those files are actual submodules defined in the same file or not is up for debate - my argument is that you usually shouldn't unless the tests are very few, because otherwise your implementation file grows unreasonably large.

> 
> I think the correct approach would be using anonymous modules that contains `discriminant`, `try_marshal_variant`, and `try_unmarshal_variant` as functions, and contains the `Marshalable` trait implementation.

I tried this and it doesn't really work. We can make it work with some trade-offs, but let's discuss this in the other PR I'll open. I'm leaning towards making a trait still, for reasons I'll explain in the PR I'll send out later.

---

Anyways, TL;DR, I wasn't trying to continue this PR, I'll split it into other PRs - I just wanted to respond to your questions is all. I'll close this for now and open new smaller PRs.

I was trying to "rip this off like a bandaid" and just get through this refactoring, but that was not easy to review, so I should have just sent smaller PRs anyways. When I wrote this I was still thinking I was blocking other work on this and wanted to just get done with it, but I've agreed in our other PRs to not be so harsh and block other work.

So it's my confusion - I'll send PRs tomorrow about these. Monday is a bit heavy of a meeting day for me, unfortunately.

Issue #68, opened 2024-04-02
Why `Marshal` and `Marshalable`?


oddcoder:
I used to see derive macros, and their respective trait sharing the same name. But the macro called `Marshal` while the trait is called `Marshalable`. Is there a particular reason or preference?

trent-reed:
I fully agree with this rename. I'll make this rename as soon as I am able.

PR #80, opened 2024-08-06
Update the Marshal proc-macro to Marshalable

trent-reed:
This will allow for a more predictable and familiar experience when working with the Marshalable proc-macro. Most derive proc-macros are simply named the same thing as the trait that they implement, so we should do the same thing here.

fixes: #68

### DIFF
--- a/base/src/commands.rs
+++ b/base/src/commands.rs
@@ -1,5 +1,5 @@
 use crate::constants::{TPM2Cap, TPM2CC, TPM2PT, TPM2SU};
-use crate::{Marshal, Marshalable, UnmarshalBuf};
+use crate::{Marshalable, UnmarshalBuf};
 use crate::{TpmiYesNo, TpmlDigest, TpmlPcrSelection, TpmsCapabilityData};
 
 /// Trait for a TPM command transaction.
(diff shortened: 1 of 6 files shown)


oddcoder (approved):
Looks good to me. While we're at it, what do you think about renaming both crates from `marshal`/`marshal-derive` to `marshalable`/`marshalable-derive` to be consistent with the rest of the renaming? Feel free to merge as-is or update the names before merging, whichever you prefer.

trent-reed:
I actually think there's a discussion here for what we want the crate topology to be. I'd rather follow-up on the other threads of cleanup before we decide to rename crates, because I suspect some crates can go away entirely. So for now, I'd like to just keep doing clean-up and see where that lands us.

Thanks for the review!

trent-reed:
> I actually think there's a discussion here for what we want the crate topology to be. I'd rather follow-up on the other threads of cleanup before we decide to rename crates, because I suspect some crates can go away entirely. So for now, I'd like to just keep doing clean-up and see where that lands us.
> 
> Thanks for the review!

Well, okay - I say this, but I guess I don't see the harm in keeping the crate names consistent for the time being. I'll send another quick PR to keep the crate consistently named, but eventually I want to explore which of these crates we need and what we want them to be named before we publish for the first time on crates.io.

Thanks!

trent-reed:
Opened https://github.com/tpm-rs/tpm-rs/pull/81 to clean this up.

PR #81, opened 2024-08-06
Rename tpm2-rs-marshal and tpm2-rs-marshal-derive

trent-reed:
A prior PR (https://github.com/tpm-rs/tpm-rs/pull/80) updated the `Marshal` proc-macro to be `Marshalable` in order to make the naming more consistent. However this creates a strange naming standard for the crate - which was originally named `tpm2-rs-marshal` and `tpm2-rs-marshal-derive`. This commit renames the crates to match with the trait and proc-macro that the crates are providing.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -5,8 +5,8 @@ members = [
     "client",
     "client-features",
     "errors",
-    "marshal",
-    "marshal-derive",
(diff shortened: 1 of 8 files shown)


oddcoder (approved):
look good, thanks!

Issue #82, opened 2024-08-07
Separate unit tests into a different file when there are many of them


trent-reed:
# Summary

When there are a lot of unit tests, we should just move the tests into a separate file instead of leaving them in-line in the module.

# Details

When you set up tests like this:

```rs
#[cfg(test)]
mod tests {
    #[test]
    fn test_1() { ... }

    #[test]
    fn test_2() { ... }
}
```

It's identical to setting up the same tests in a separate submodule file named `tests.rs` (or `tests/mod.rs`):

```rs
#[cfg(test)]
mod tests;
```

```rs
#[test]
fn test_1() { ... }

#[test]
fn test_2() { ... }
```

It's actually a property that is left up to the crates themselves to decide how they do it.

Even though it's fairly common to see rust projects that just define all of the tests in the same file, I've never seen good justification for that, especially in cases where there are a lot of tests.

* It's easier to read either just the tests or just the implementation when they are separate from one-another.
* It doesn't harm comparison of tests to implementation because one can just open both files side-by-side to view them both at the same time (and one would anyways have to do this even if the tests were in the same file, since the tests are usually way at the bottom of a file).
* It's easier to see where the tests are because you can just look for submodules named `tests` on the filesystem.
* Tests are code too, and I think we should usually follow similar strategies for clean code organization, that way we encourage developers to treat tests as similarly impactful as the code that it operates on (to keep things clean and well-tested).
* It adds extra indentation that can be annoying especially if you have helper types near your tests, plus if you're testing thoroughly it at least doubles (if not more) the length of the file. Just seems more readable in a separate file.

# A Note on the Separate `tests/` Directory Outside of `src/`

This directory acts kind of special - it's handled by treating each `*.rs` file as a separate test-only crate that can be run as a set of tests and depends on your main crate. So the difference between putting tests in `src/` versus `tests/` is basically: Do you want to be testing code as-if a client would be using it - or as it's more commonly put, integration tests.

My take is that unit tests should be close to the code they test, but in submodules when the number of tests is high. Personally I'd like to always see these in separate files, but that may be annoying in some cases with very little logic, so I'm willing to bend in some cases.

[...]

PR #83, opened 2024-08-07
Separate unit tests into a different file when there are many of them

trent-reed:
This change will make it easier to manage our unit tests.

There is nothing special about the `test` module in rust libraries, it's just a separate module that is conditionally compiled based on whether or not the crate is running with the `test` command. As such, it's possible (and allowable) to separate the unit test code into a separate file.

It would probably be wise to move these tests closer to the code that they test, but let's do that in the future.

fixes: #82


trent-reed:
Oh, forgot to mention, this is the last of my changes that I had already started on this repo; so after this I'll be moving to my fork. Just a quick FYI on that.

client/tests/integration_tests.rs
```diff
@@ -0,0 +1,13 @@
...
+/// command line.
+///
+/// To run locally in the docker container:
+///  `cd client && docker compose run simulator_tests`
+///
+/// These tests must be run with `--test-threads=1`, because they use a single TCP port.
+
+// Separate module to avoid us running these tests under normal `cargo test` runs.
+#[cfg(feature = "tpm-simulator-tests")]
+mod simulator_tests;
```
oddcoder, line 13: can't we just move `client/tests/simulator_tests/mod.rs` to `client/tests/simulator_tests.rs` place a 
`#![cfg(feature = "tpm-simulator-tests")]` on line 1, delete everything else in this directory and it should still work exactly the same, with less nested structure.

oddcoder: Amusing that actually works, we might want to keep the comments of this file in `client/tests/simulator_tests.rs`

oddcoder (approved):
everything else looks good to me.

> Oh, forgot to mention, this is the last of my changes that I had already started on this repo; so after this I'll be moving to my fork. Just a quick FYI on that.

Thanks for letting me know, I didn't actually notice until you mentioned it, my work flow is usually something along the lines:

```
wget https://github.com/tpm-rs/tpm-rs/pull/83/commits/310d60f7bc9778a703af5c5226bec987d4dbcea8.patch
```
perhaps that is why I always mix the words commit/PR/patch.

I made little suggestion, if it work, it might be nice to remove the unnecessary file. If not please ignore it. I already approved to avoid blocking you till next day.

client/tests/integration_tests.rs
trent-reed: Ah yeah sorry, I didn't mention this in the PR description, but I did actually already try this and it doesn't work. :(

trent-reed: Oh wait, it might work, it might have failed because I had it after the doc comment and that is a fatal error - I didn't notice that. Let me try again.

trent-reed: Yup, it works! Updated the PR - I'll squash and merge. Thank you so much for the quick review!

Issue #84, opened 2024-08-07
Add a New Trait: `MarshalableVariant`


trent-reed:
# Summary

The `Marshalable` crate has some private functions that leak into being accessible to other types in the case of `enum` items currently. I propose that we officially say that these functions can be accessed from the public API, and say they are under a new trait named `MarshalableEnum`.

# Problem Statement

Today, we have the following trait:

```rs
pub trait Marshalable {
    fn try_unmarshal(buffer: &mut UnmarshalBuf) -> TpmRcResult<Self>
    where
        Self: Sized;
    fn try_marshal(&self, buffer: &mut [u8]) -> TpmRcResult<usize>;
}
```

There's actually a few issues here (for example, `TpmRcResult` is not the right error type; it's actually supposed to be a different error type for client vs server - we can solve this with an indirect error type; I'll open a separate issue for that), but for now let's just focus on what this issue is talking about - the accidentally visible enum functions.

When implementing `Marshalable` for an enum, the proc-macro define the following private functions on the type:

```rs
impl $Type {
    fn discriminant(&self) -> #Primitive;
    fn try_marshal_variant(&self, buffer: &mut [u8]) -> TpmRcResult<usize>;
    fn try_unmarshal_variant(selector: #Primitive, buffer: &mut UnmarshalBuf) -> TpmRcResult<Self>;
}
```

Now this is okay so far, it makes sense to separate the functions out, no harm no foul.

Where things become questionable is when you have another separate type that uses the enum, and does ***NOT*** have the discriminant placed right before the enum data.

Like this:

```rs
pub struct TpmsAttest {
    pub magic: TPM2Generated,
    // NOTE: There is a hidden discriminant here!
    pub qualified_signer: Tpm2bName,
    pub extra_data: Tpm2bData,
    pub clock_info: TpmsClockInfo,
    pub firmware_version: u64,
    pub attested: TpmuAttest, // This is the enum for the discriminant!
}
```

This creates a few problems:

1. It currently makes it impossible for us to mark this up and use the `Marshalable` proc-macro in a sensible way.
2. Since these functions are private, it's impossible to define an enum in one module, and a structure with an out-of-line discriminant for that enum in a different module (because the functions are private).
3. By extension, if a vendor wishes to provide custom types/requests/responses/commands (something that I understand is actually quite common in TPMs), then this format currently does not work cross-crate.

[...]

PR #89, opened 2024-08-14
Add a new trait: MarshalableVariant

trent-reed:
This change starts the process of implementing a separate trait for marshaling variant data independently of the discriminant. This PR purposefully aims to do the minimal amount of changes so that we can unblock other cleanup in other PRs.

After this PR we will do more cleanup in the repo, and eventually we will separate this out into its own derive proc-macro (and add more thorough unit tests). But to unblock the cleanup of tpm2-rs-base, we need to start with this.

See the discussion about this: https://github.com/tpm-rs/tpm-rs/issues/84

Changes:

*   Added a new trait named `MarshalableVariant`.
*   Added a single test in `tpm2-rs-marshalable`, we can do better here but I'd like to refactor the error codes first, because right now it's hard to work with. So for now, since I'm not making any functional changes, I'd like to do the bare minimum in testing (we'll add more later).

### DIFF
--- a/marshalable/src/lib.rs
+++ b/marshalable/src/lib.rs
@@ -34,6 +34,49 @@ pub trait Marshalable {
     fn try_marshal(&self, buffer: &mut [u8]) -> TpmRcResult<usize>;
 }
 
+/// Defines the ability to marshal an enum by its variant data alone.
+///
+/// The separation of this trait from `Marshalable` is useful for cases where
+/// an enum's data is not directly after the variant's selector. This is
+/// something that can happen in some TPM types, so we need to be able to
+/// support it (and it needs to be able to function cross-crates in order to
+/// support vendors creating their own types, commands, etc).
+///
+/// # Notice
+///
+/// Today, you can get this trait simply by deriving the `Marshalable` trait.
+/// However, in the future we will be separating this out into a different
+/// derive proc-macro.
+///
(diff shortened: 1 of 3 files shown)


marshalable/src/lib.rs
```diff
@@ -34,6 +34,49 @@ pub trait Marshalable {
...
+    ///
+    /// Because the way the data in `buffer` is interpreted changes depending on
+    /// the variant we are unmarshaling, we have to be told explicitly which
+    /// variant is being targeted.
+    fn try_unmarshal_variant(
+        selector: Self::Selector,
+        buffer: &mut UnmarshalBuf,
+    ) -> TpmRcResult<Self>
+    where
+        Self: Sized;
```
oddcoder, line 69: Optional Nit: `Where Self: Sized` is the same as ` pub trait MarshalableVariant: Sized`  But the later looks simpler.

oddcoder (approved):
aside from 1 small nit, everything looks good, thanks!

marshalable/src/lib.rs
trent-reed: I suppose since `MarshalableVariant` deals specifically with enums, one could make the argument for `MarshalableVariant` having a supertrait of `Sized` even if `Marshalable` doesn't...

Anyways, probably best to discuss this in a separate PR, I'll send one out after merging. Thanks!

PR #90, opened 2024-08-15
Change to using a supertrait for Marshalable traits

trent-reed:
Note that this does say something subtly different than what the original code said. The original code was saying that one could implement `Marshalable` or `MarshalableVariant` even if the type it was being implemented on wasn't `Sized` (like a slice, for example).

This would have allowed one to implement these traits on dynamically sized types (DSTs). However, the utility of this is not clear, and it seems simpler to just say that if you can marshal something, you should also be able to unmarshal it.


oddcoder (approved):
look good, thanks!

PR #91, opened 2024-08-15
Small patch to fix `UnmarshalBuf` to be fully-qualified

trent-reed:
Small patch to fix `UnmarshalBuf` to be fully-qualified.

### DIFF
--- a/marshalable-derive/src/lib.rs
+++ b/marshalable-derive/src/lib.rs
@@ -66,21 +66,24 @@ fn derive_tpm_marshal_inner(input: DeriveInput) -> Result<TokenStream> {
             ));
         }
     };
-
     let expanded = quote! {
-        #pure_impl
-        // The generated impl.
-        impl Marshalable for #name  {
-            fn try_unmarshal(buffer: &mut UnmarshalBuf) -> tpm2_rs_marshalable::exports::errors::TpmRcResult<Self> {
-                #unmarsh_text
-            }
+        #[doc(hidden)]
+        const _: () = {
+            use tpm2_rs_marshalable::__prelude::*;
+            #pure_impl
+            // The generated impl.
+            impl Marshalable for #name  {
+                fn try_unmarshal(buffer: &mut UnmarshalBuf) -> TpmRcResult<Self> {
+                    #unmarsh_text
+                }
 
-            fn try_marshal(&self, buffer: &mut [u8]) -> tpm2_rs_marshalable::exports::errors::TpmRcResult<usize> {
-                let mut written: usize = 0;
-                #marsh_text;
-                Ok(written)
+                fn try_marshal(&self, buffer: &mut [u8]) -> TpmRcResult<usize> {
+                    let mut written: usize = 0;
+                    #marsh_text;
+                    Ok(written)
+                }
             }
-        }
+        };
     };
 
     Ok(expanded)
--- a/marshalable-derive/src/lib.rs
+++ b/marshalable-derive/src/lib.rs
@@ -67,21 +67,19 @@ fn derive_tpm_marshal_inner(input: DeriveInput) -> Result<TokenStream> {
         }
     };
     let expanded = quote! {
-        const _: () = {
-            #pure_impl
-            // The generated impl.
-            impl Marshalable for #name  {
-                fn try_unmarshal(buffer: &mut tpm2_rs_marshalable::UnmarshalBuf) -> tpm2_rs_marshalable::__private::TpmRcResult<Self> {
-                    #unmarsh_text
-                }
-
-                fn try_marshal(&self, buffer: &mut [u8]) -> tpm2_rs_marshalable::__private::TpmRcResult<usize> {
-                    let mut written: usize = 0;
-                    #marsh_text;
-                    Ok(written)
-                }
+        #pure_impl
+        // The generated impl.
+        impl Marshalable for #name  {
+            fn try_unmarshal(buffer: &mut tpm2_rs_marshalable::UnmarshalBuf) -> tpm2_rs_marshalable::__private::TpmRcResult<Self> {
+                #unmarsh_text
             }
-        };
+
+            fn try_marshal(&self, buffer: &mut [u8]) -> tpm2_rs_marshalable::__private::TpmRcResult<usize> {
+                let mut written: usize = 0;
+                #marsh_text;
+                Ok(written)
+            }
+        }
     };
 
     Ok(expanded)
--- a/marshalable-derive/src/lib.rs
+++ b/marshalable-derive/src/lib.rs
@@ -69,7 +69,7 @@ fn derive_tpm_marshal_inner(input: DeriveInput) -> Result<TokenStream> {
     let expanded = quote! {
         #[doc(hidden)]
         const _: () = {
-            use tpm2_rs_marshalable::__prelude::*;
+            use tpm2_rs_marshalable::__private::*;
             #pure_impl
             // The generated impl.
             impl Marshalable for #name  {
--- a/marshalable-derive/src/lib.rs
+++ b/marshalable-derive/src/lib.rs
@@ -67,17 +67,15 @@ fn derive_tpm_marshal_inner(input: DeriveInput) -> Result<TokenStream> {
         }
     };
     let expanded = quote! {
-        #[doc(hidden)]
         const _: () = {
-            use tpm2_rs_marshalable::__private::*;
             #pure_impl
             // The generated impl.
             impl Marshalable for #name  {
-                fn try_unmarshal(buffer: &mut UnmarshalBuf) -> TpmRcResult<Self> {
+                fn try_unmarshal(buffer: &mut tpm2_rs_marshalable::UnmarshalBuf) -> tpm2_rs_marshalable::__private::TpmRcResult<Self> {
                     #unmarsh_text
                 }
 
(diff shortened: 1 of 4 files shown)


marshalable/src/lib.rs
```diff
@@ -8,7 +8,7 @@ pub use tpm2_rs_marshalable_derive::Marshalable;
 // This is a module that is exposed publicly, but isn't expected to be used.
 // It's for the internal usage of having a prelude for the marshalalbe macro.
 #[doc(hidden)]
-pub mod __prelude {
+pub mod __private {
```
trent-reed, line 11: I ended up going with this name to more strongly discourage someone from using an import from it. This is the same name that serde uses for their re-exports: https://github.com/serde-rs/serde/blob/1a9ffdbd0c1d4d8dd4f7e686e00747c932a36d3d/serde/src/lib.rs#L316

marshalable-derive/src/lib.rs
```diff
@@ -66,21 +66,24 @@ fn derive_tpm_marshal_inner(input: DeriveInput) -> Result<TokenStream> {
...
     let expanded = quote! {
-        #pure_impl
-        // The generated impl.
-        impl Marshalable for #name  {
-            fn try_unmarshal(buffer: &mut UnmarshalBuf) -> tpm2_rs_marshalable::exports::errors::TpmRcResult<Self> {
-                #unmarsh_text
-            }
+        #[doc(hidden)]
+        const _: () = {
+            use tpm2_rs_marshalable::__private::*;
```
oddcoder, line 72: the #[doc(hidden)], anon module are not necessary. because you are trying to hide the fact that a trait is being implemented (at last that what code feel like it is trying). But that is not possible in rust.

As for the `use` I think it might not be necessary if we stick to the rule of fully qualified names

marshalable-derive/src/lib.rs
```diff
@@ -66,21 +66,24 @@ fn derive_tpm_marshal_inner(input: DeriveInput) -> Result<TokenStream> {
...
+            impl Marshalable for #name  {
+                fn try_unmarshal(buffer: &mut UnmarshalBuf) -> TpmRcResult<Self> {
+                    #unmarsh_text
+                }
 
-            fn try_marshal(&self, buffer: &mut [u8]) -> tpm2_rs_marshalable::exports::errors::TpmRcResult<usize> {
-                let mut written: usize = 0;
-                #marsh_text;
-                Ok(written)
+                fn try_marshal(&self, buffer: &mut [u8]) -> TpmRcResult<usize> {
```
oddcoder, line 80: The reason I prefer fully qualified names over uses is because it makes it much much easier to debug when full path is not provided, but with `use`, it i a bit harder and it requires reviewer to look at the final macro output to see if there is a missing `use`.

marshalable/src/lib.rs
oddcoder: I am not sure if this works the way you intended. Everything is still visible in documentation

So first I think `Marshalable` and all types referenced by `Marshalable` to be fully public, otherwise users will not be either  able to {un}marrshal.

As for `MarshalableVariant` it would b ok to put it in the hidden module. But It should be defined in that module and not in top level otherwise it will always be visable (I think ... I didn't really test it).

trent-reed: > I am not sure if this works the way you intended. Everything is still visible in documentation

This is working how I intended.

Without this change, you can see the following:

image

With the change you cannot.

---

> So first I think Marshalable and all types referenced by Marshalable to be fully public, otherwise users will not be either able to {un}marrshal.

I don't have a problem with them being `pub`, my issue is leaving the module open for folks to take a direct reference on. If we name it something like `exports` that feels to me like it's not pushing the client hard enough to not take a direct dependency on it.

I don't want them to, I want the things that are exported here to be an implementation detail of `Marshalable`. I know they *need* the dependency, but I'd rather urge them to go through the proper path to get that dependency instead of reaching into the guts of `__private` to scrape them out; I don't want them to take a dependency on the names in here.

I'd like to imply more firmly that this is an implementation detail of `Marshalable` and friends.

---

> As for MarshalableVariant it would b ok to put it in the hidden module. But It should be defined in that module and not in top level otherwise it will always be visable (I think ... I didn't really test it).

It made more sense to put it in the hidden module if we were treating it like a prelude. What I really want to fix here is:

* hide the module so that folks can't see it in documentation
* name it something that discourages clients linking against the crate from using things exported from this module
* fix some situations where we accidentally had dependencies provided by the caller's context (like the `UnmarshalBuf`, I think that's the only instance of it).

marshalable-derive/src/lib.rs
trent-reed: I did that because serde did it and didn't question it, but you're right we don't need it (and likely they don't either - it came from a time before they used `_` to name these code blocks).

I'm not trying to hide the trait implementation, I'm trying to hide the const block, but on second look that seems unneeded.

trent-reed: Apparently blames to this: https://github.com/serde-rs/serde/commit/a135199ab104a8068f6dd672f813760d17128212

Anyways, I'll remove this.

marshalable-derive/src/lib.rs
trent-reed: I find the long names hard to read; but I don't mind either way, this isn't something I mind just leaving as-is; it felt clean to me, but if you disagree I'll change it back.

oddcoder: I slightly prefer fully qualified names.

marshalable-derive/src/lib.rs
```diff
@@ -66,21 +66,22 @@ fn derive_tpm_marshal_inner(input: DeriveInput) -> Result<TokenStream> {
...
     };
-
     let expanded = quote! {
-        #pure_impl
-        // The generated impl.
-        impl Marshalable for #name  {
-            fn try_unmarshal(buffer: &mut UnmarshalBuf) -> tpm2_rs_marshalable::exports::errors::TpmRcResult<Self> {
-                #unmarsh_text
-            }
+        const _: () = {
```
oddcoder, line 70: the anonymous module is not necessary, or am I missing something?

marshalable/src/lib.rs
oddcoder: I am confused, reading my comment gave me the the impression that there was `pub use MarshalableVariant` But the code does not look like that, I am not sure if you changed it or it is like this all the time.


I slightly prefer one of two things:
1- public re-export of `TpmRcError` and `TpmRcResult`
2- requiring users to add tpm2_rs_errors to `cargo.toml`

I prefer `1` over `2`, but I prefer both over the current approach. keep in mind users may have to implement `Marshalable` manually so they need access to `TpmRcError` and `TpmRcResult`

That said, I think `MarshalableVariant` should be in `__private`, because no one should have access to it.

marshalable-derive/src/lib.rs
trent-reed: I suppose it's no longer necessary, I can remove it - it would protect us from future refactors adding helper functions and accidentally exposing them again, but we currently aren't creating any more of such functions.

I can remove it.

marshalable/src/lib.rs
trent-reed: The code was updated to no longer have types defined by `tpm2-rs-marshalable` in the `__private` modules, I agree that it was not necessary to re-export things in `__private` that `tpm2-rs-marshalable` itself defined.

My original aim here was:

* You don't need to worry about the implementation of `Marshalable` or `MarshalableVariant`, the proc-macro can implement the trait regardless of what your current code has `use` statements for.
* If you want to do something explicit with types like `TpmRcError`, etc. You have to add an explicit `use` statement for them, and we prefer if you don't re-use the `use` statements in `__private` of `tpm2-rs-marshalable` (since those are for implementation detail). Thus, the dev should `use` from their own `cargo.toml` dependency of `tpm2-rs-errors`.

This does actually still allow one to manually implement `Marshalable` and friends, but it creates a requirement that the version of the `tpm2-rs-errors` crate that their `cargo.toml` links in is the same as the one used by `tpm2-rs-marshalable`, which might be a tad annoying instead of just re-exporting.

From a cursory glance around forums this is another thing that nobody really agrees on in the rust community. Some folks say to re-export the types, others say to just re-export the crate itself if your crate's public API uses a type from the given crate, few seem to say to avoid cross-crate re-export (because of that same version thing I mentioned, it seems).

It seems like there's a stronger preference for just re-exporting the crates themselves (something like `pub extern crate foo;`), what do you think of that instead?

I think that's quite nice, and it makes it clearer that it's a cross-crate dependency. It looks like this in the generated documentation:

image

---

> That said, I think `MarshalableVariant` should be in `__private`, because no one should have access to it.

Both `Marshalable` and `MarshalableVariant` need to be accessed cross-crate. That's the whole point of making `MarshalableVariant` a trait on its own, otherwise they would have stayed private functions - so that we can use the functionality cross-crate for cases where the discriminant is not directly before the variant data (which unfortunately is a case that pops up in the TPM specification).

I **do** want the `MarshalableVariant` trait to be both used, and visible publicly from this crate.

trent-reed: You know what I just realized, none of this matters because we're removing the only cross-crate dependency we have in `exports` via issue 92.

I'll just revert this part and we'll leave it as is for now.

oddcoder (approved):
Looks good, thanks!

Issue #100, opened 2024-08-29
Rename `TpmError` to `TssError`


trent-reed:
## Summary

The type `TpmError` is trying to represent any `TSS` error (that is to say, an error from a TPM, or some other layer of the `TSS` stack). However it's currently confusingly named `TpmError`, which makes it sound like it's related specifically to a `TPM` error.

## Details

* Rename `TpmError` -> `TssError`
* Rename `TpmResult` -> `TssResult`

PR #101, opened 2024-08-29
Rename `TpmError` to `TssError`

trent-reed:
These types are trying to talk about `TSS` errors, but they are named `TPM` errors which is quite confusing. This rename should help clarify things.

* Rename `TpmError` to `TssError`
* Rename `TpmResult` to `TssResult`

FIXES: #100

### DIFF
--- a/client-features/src/lib.rs
+++ b/client-features/src/lib.rs
@@ -1,7 +1,7 @@
 #![forbid(unsafe_code)]
 use tpm2_rs_base::commands::*;
 use tpm2_rs_base::constants::{TPM2Cap, TPM2PT};
-use tpm2_rs_base::errors::{TpmRcError, TpmResult};
+use tpm2_rs_base::errors::{TpmRcError, TssResult};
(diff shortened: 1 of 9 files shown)


oddcoder:
I am confused,  why are we doing this?

trent-reed:
I think it's confusing to have two different error types, one named `TpmRcError` and one named `TpmError`. The name `TpmError` implies heavily that it's a TPM error, but it's not, it's a TSS error (which may be but doesn't have to be a TPM error).

I think the naming is wrong, so I'm proposing we change it.

I'll add an issue this fixes to the description.

TRJensen (approved):
Agree - TpmRcError reflects the possible error codes from TPM spec - whereas the other TssResult reflects other types of errors.

trent-reed:
Thanks for the +1 @TRJensen

I'm gonna go ahead with this change. I see a future where we use these TSS errors in fewer places, but for now I just want to avoid the confusion of the type name.

PR #104, opened 2024-09-04
Remove client-features folder and content.

### LINKED ISSUE #103: Propose to cleanup of client-features folder
The client-features folder contains a single wrapper function `get_manufacturer_id`, which creates a `GetCapabilityCmd` struct and sends the request via `run_command`.

Based on discussions in https://github.com/tpm-rs/tpm-rs/pull/63, we should avoid having function wrappers for TPM api interface, and instead let user create create the Cmd structure, and call the run_command() - or run_command_with_handles().

I would suggest to:
- Remove "client-features" folder (incl get_manufacturer_id function).
- Adapt and move test vectors to the client folder.

TRJensen:
Removal of client-features folder, to avoid having function wrappers for TPM API interface. The user should instead create the Cmd structure, and call the run_command() - or run_command_with_handles().

FIXES #103


TRJensen:

oddcoder:
Hello @TRJensen, Thanks for the PR,  I think the general idea is fine 

Please note that tmp emulator tests did not pass. Those tests do not run directly when cargo test and should be invoked manually.

trent-reed:
Yes, @oddcoder is right, the tests don't pass (they don't build).

If you're adding a simulator test, you need to manually invoke them, because they involve spinning up a simulator instance by building a Docker container. That takes a bit more time than usual unit tests with mocking, so it's usually a manually-done thing.

Note that client-features/src/tests.rs was a unit test, not an integration test, where as client/tests/simulator_tests.rs is an integration test, I'm not sure if you intended on making this change.

At the top of the file, there's a comment with the command needed to run the simulator tests: `cd client && docker compose run simulator_tests`

Please get these tests passing and then we will look at the PR again.

client/tests/simulator_tests.rs
```diff
@@ -51,9 +50,18 @@ fn get_started_tpm() -> (TpmSim, TcpTpm) {
...
     let (_sim_lifeline, mut tpm) = get_started_tpm();
-    assert!(get_manufacturer_id(&mut tpm).is_ok());
+
+    const CMD: GetCapabilityCmd = GetCapabilityCmd {
+        capability: TPM2Cap::TPMProperties,
+        property: TPM2PT::Manufacturer,
+        property_count: 1,
+    };
+    let resp = run_command(&CMD, tpm)?;
+    assert!(let TpmsCapabilityData::TpmProperties(prop) = resp.capability_data);
```
trent-reed, line 62: I don't think you can do a `let` inside an assert like this... Though I've never tried it myself, I do see that this is causing a build error for you.

client/tests/simulator_tests.rs
```diff
@@ -51,9 +50,18 @@ fn get_started_tpm() -> (TpmSim, TcpTpm) {
...
+    assert!(prop.count() == 1);
```
trent-reed, line 63: I would prefer if you used `assert_eq!` in the cases where it's possible to use.

client/tests/simulator_tests.rs
trent-reed: The way to do this btw is like this:

```rs
let TpmsCapabilityData::TpmProperties(prop) = resp.capability_data else { panic!("Unexpected variant data.") };
```

This is okay for testing, but you should not do such a thing in non-test code.

client/tests/simulator_tests.rs
trent-reed: In fact, if possible (and stable to do), best case would be that you set up an explicit expectation structure, and just check equality against that:

```rs
let mut expected = TpmlTaggedTpmProperty {
    count: 1,
    tpm_property: [TpmsTaggedProperty::default(); TPM2_MAX_TPM_PROPERTIES],
};
expected.tpm_property[0] = TpmsTaggedProperty {
    ...,
};
assert_eq!(resp.capability_data, TpmsCapabilityData::TpmProperties(expected));
```

You'd need to add `PartialEq` on `TpmsCapabilityData` for this, which I think is reasonable, personally.

If you want to avoid that, you could just check equality over the one property doing what you're doing today (because `TpmlTaggedTpmProperty` derives `PartialEq`):

```rs
let TpmsCapabilityData::TpmProperties(props) = resp.capability_data else { panic!("Unexpected variant data.", ) };
let mut expected = TpmlTaggedTpmProperty {
    count: 1,
    tpm_property: [TpmsTaggedProperty::default(); TPM2_MAX_TPM_PROPERTIES],
};
expected.tpm_property[0] = TpmsTaggedProperty {
    ...,
};
assert_eq!(props, expected);
```

---

I wish rust had better partial-initialization syntax, but it doesn't... If we wanted to make partial initialization cleaner, we could write a testonly function like this:

```rs
fn partial_initialized_array<T: Copy + Default, const N: usize>(
    init: &[T]
) -> [T; N] {
    let mut arr = [T::default(); N];
    arr[..init.len()].copy_from_slice(init);
    arr
}
```

Then our `expected` can be a lot cleaner represented:

```rs
let expected = TpmlTaggedTpmProperty {
    count: 1,
    tpm_property: partial_initialized_array(&[
        TpmsTaggedProperty{ ... },
    ]),
};
```

TRJensen:
@oddcoder and @trent-reed - Thank you for the great input :slightly_smiling_face:

I wanted to adjust the test cases after removing the clients-features folder. Had to run for another appointment and quickly changed to "draft", before being fully able to investigate further - and later busy with other activities. I should probably have waited adding reviewers, so sorry for the noise :face_with_diagonal_mouth:. But got some great input :+1: Will have a look...

client/tests/simulator_tests.rs
```diff
@@ -51,9 +52,23 @@ fn get_started_tpm() -> (TpmSim, TcpTpm) {
 }
 
 #[test]
-fn test_get_manufacturer_id() {
+fn test_get_capability_manufacturer_id() {
     let (_sim_lifeline, mut tpm) = get_started_tpm();
-    assert!(get_manufacturer_id(&mut tpm).is_ok());
+
+    const CMD: GetCapabilityCmd = GetCapabilityCmd {
```
oddcoder, line 58: is there a reason to not do the normal

```rust
let cmd =  GetCapabilityCmd {
....
```

This is fairly standard, and usually `const` is reservered for global constants

client/tests/simulator_tests.rs
TRJensen: It is a great idea - and a clean way to do a full comparison of the full struct in one go.

I did however have to make the elements of TpmlTaggedTpmProperty public, as I was not allowed to initialize them.

This may raise the question, if it is possible to make some elements public for testing purpose only, but not for a normal build - but I am not sure if there is a design strategy to keep many of struct elements private.

client/tests/simulator_tests.rs
TRJensen: Fixed

oddcoder (approved):
looks good, thanks!

base/src/lib.rs
```diff
@@ -1189,9 +1189,9 @@ pub struct TpmlCc {
 #[repr(C)]
 #[derive(Clone, Copy, PartialEq, Debug, Marshalable)]
 pub struct TpmlTaggedTpmProperty {
-    count: u32,
+    pub count: u32,
```
trent-reed, line 1192: This was actually a bug that these weren't `pub`, so good fix!

client/tests/simulator_tests.rs
```diff
@@ -51,9 +52,35 @@ fn get_started_tpm() -> (TpmSim, TcpTpm) {
...
+        count: 1,
+        tpm_property: [TpmsTaggedProperty::default(); 127],
+    };
+
+    expected.tpm_property[0] = TpmsTaggedProperty {
+        property: TpmPt::Manufacturer,
+        value: 0x4D534654,
+    };
+
+    let command: GetCapabilityCmd = GetCapabilityCmd {
```
trent-reed, line 68: Nit: You don't need to re-state `GetCapabilityCmd` here, the type is clearly known by the rhs of the assignment operator.

I'd just:

```rs
let command = GetCapabilityCmd {
    ...
};
```

trent-reed:
Mostly LGTM, just one nit - thanks!

trent-reed (approved):

TRJensen:
Thank you @oddcoder and @trent-reed for the approval - @trent-reed I just fixed the last "nit" and will go ahead and push my very first tpm-rs change in. Thanks 🙂 🎆 👍 🍾

Issue #98, opened 2024-08-28
RFC: Suggested Renames for Constant Types


trent-reed:
## Summary

There are a few types that we define today that I would like to rename. I'm requesting comments on this, to see if others agree with me or not. (For this issue, I will focus solely on types under the `constants` module.)

### Notes

*   This is happening somewhatly independently of [Issue #97](https://github.com/tpm-rs/tpm-rs/issues/97), but it may overlap in some small ways. For example, I argue in that issue that these enums are actual "types" so they don't really make sense to be under a module named `constants`

### Rename the constants

I'm not really seeing why we prefix these with `TPM2_*`, when it seems that the TPM spec version isn't included in any of the other constants in the spec names. I have a slight preference for just calling these `TPM_*` instead. I think that aligns better with what the spec names their constants.

#### High-Level Thoughts on the Constants

From [Issue #97](https://github.com/tpm-rs/tpm-rs/issues/97), you can see my feelings on enum types. However, I'll copy my statements here for visibility:

> I highly suspect that these are only named "constants" in the spec because the folks who penned the spec have the perspective of C programmers.
> 
> Enums are constants, according to C. So it makes some kind os sense to say "these types are constants". But from the perspective of rust, they're just `enum` types (well, because of `open_enum`, they're structures with constant `impl` values that look like `enums`, but let's not get too in the weeds over that).
> 
> My feeling is these should be treated like other types. So they should be named similarly (`PascalCase`), and if we decide to organize types into a module named `types`, they actually belong there - not in the `constants` module.

There's one thing I don't talk about in this text blurb; The types in the spec are named things like `TPM_CAP`, and `TPM_ST`. So I'm not sure why we named these types `TPM2*` when none of the names have a 2 present in them. I also will be suggesting that we remove this from the names.

Also, the proper naming format would be `UpperPascalCase` for enum types, so I'd like to rename to that format instead of leaving them all caps.

#### About `UpperPascalCase` and Acronyms

One thing you invariably run into when naming things in `UpperPascalCase` is what to do with acronyms.

In my experience, the best thing to do is to **NOT** keep the acronyms all-caps. The reason for this is if you have two acronyms next to each other, you need to introduce something to split the words, otherwise they blend together.

So for example, I wouldn't want to name these types `TPMCC`, I'd want to name it `TpmCc` (Similar for other instances of non-constant names in the codebase, like `SHA` should be `Sha`, and so on).

#### About the Actual Target Name

[...]

PR #105, opened 2024-09-18
Rename enum types to be formatted in `UpperPascalCase`

trent-reed:
This change updates the typenames to be formatted as appropriate for rust enum types. It also removes the prefixing of `TPM2`, instead opting for just `Tpm` (most of these types were originally just named `TPM_*`, so I think this is appropriate).

One other minor changes fell out of this:

*   There's no longer a `TpmCc` and a `Command` alias, we are
    canonicalizing to the formal name from the spec (`TpmCc`).

FIXES: #98

### DIFF
--- a/base/src/commands.rs
+++ b/base/src/commands.rs
@@ -1,11 +1,11 @@
-use crate::constants::{TPM2Cap, TPM2CC, TPM2PT, TPM2SU};
+use crate::constants::{TpmCap, TpmCc, TpmPt, TpmSu};
 use crate::Marshalable;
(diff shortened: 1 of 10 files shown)


trent-reed:
Sorry to keep adding you to all my reviews @oddcoder, but we've worked the most together, and you commented on the issue that I'm proposing this fix for. :)

oddcoder (approved):
looks good, thanks!

> There's no longer a `TpmCc` and a `Command` alias, we are
    canonicalizing to the formal name from the spec (`TpmCc`).


I am also in favor of this, thanks!

PR #108, opened 2024-09-30
Make the link clickable in docs and add CI to catch documentation issues

wiktor-k:
Btw the CI uses `actions-rs/cargo` action which, according to the repo, is deprecated: https://github.com/actions-rs/cargo

I think converting to plain cargo invocations would be a good idea (but I didn't want to spam this PR with too many changes at once).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -78,3 +78,19 @@ jobs:
         with:
           command: clippy
           args: --all-targets -- -D warnings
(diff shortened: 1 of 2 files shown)


wiktor-k:

oddcoder (approved):
Thanks for doc fix and CI integration, looks good, feel free to send another PR using CLI instead of actions-rs if you would like to!

PR #110, opened 2024-10-02
Add guiding comments to "commands.rs" to match [TPM2.0 1.83]

### LINKED ISSUE #109: Suggestion to add TPM command spec references into "commands.rs" file.
Idea is simply to refer to latest TPM 2.0 1.83 Part 3 Commands specification.

For example:

<pre>
// ============================================================================
// 32 Attached Components [TPM2.0 1.83]
// ============================================================================

// 32.2 TPM2_AC_GetCapability

&lt; ... request and response struct here ... &gt;

// 32.3 TPM2_AC_Send

// 32.4 TPM2_Policy_AC_SendSelect
</pre>

This helps keeping command request and response definitions sorted as pr specification, and make it easy to search definition based in naming from specification.

TRJensen:
Comments with chapters and subsections added to "commands.rs", so that it matches sections in TPM-2.0-1.83 Part 3 Commands specification.

FIXES #109

### DIFF
--- a/base/src/commands.rs
+++ b/base/src/commands.rs
@@ -1,6 +1,105 @@
-use crate::constants::{TpmCap, TpmCc, TpmPt, TpmSu};
+use crate::constants::TpmCc;
 use crate::Marshalable;
-use crate::{TpmiYesNo, TpmlDigest, TpmlPcrSelection, TpmsCapabilityData};
+
+// [TPM2.0 1.83] 9 Start-up
+mod startup;
+pub use startup::*;
+
+// [TPM2.0 1.83] 10 Testing
+mod testing;
+pub use testing::*;
+
+// [TPM2.0 1.83] 11 Session Commands
+mod session;
+pub use session::*;
+
+// [TPM2.0 1.83] 12 Object Commands
+mod object;
+pub use object::*;
+
+// [TPM2.0 1.83] 13 Duplication Commands
+mod duplication;
+pub use duplication::*;
+
+// [TPM2.0 1.83] 14 Asymmetric Primitives
+mod asymmetric;
+pub use asymmetric::*;
+
+// [TPM2.0 1.83] 15 Symmetric Primitives
+mod symmetric;
+pub use symmetric::*;
+
+// [TPM2.0 1.83] 16 Random Number Generator
+mod random;
+pub use random::*;
+
+// [TPM2.0 1.83] 17 Hash/HMAC/Event Sequences
+mod hash_hmac_event;
+pub use hash_hmac_event::*;
+
+// [TPM2.0 1.83] 18 Attestation Commands
+mod attestation;
+pub use attestation::*;
+
+// [TPM2.0 1.83] 19 Ephemeral EC Keys
+mod ephemeral;
+pub use ephemeral::*;
+
+// [TPM2.0 1.83] 20 Signing and Signature Verification
+mod signature;
+pub use signature::*;
+
+// [TPM2.0 1.83] 21 Command Audit
+mod command_audit;
+pub use command_audit::*;
+
+// [TPM2.0 1.83] 22 Integrity Collection (PCR)
+mod pcr;
+pub use pcr::*;
+
+// [TPM2.0 1.83] 23 Enhanced Authorization (EA) Commands
+mod enhanced_auth;
+pub use enhanced_auth::*;
+
+// [TPM2.0 1.83] 24 Hierarchy Commands
+mod hierarchy;
+pub use hierarchy::*;
+
+// [TPM2.0 1.83] 25 Dictionary Attack Functions
+mod dictionary_attack;
+pub use dictionary_attack::*;
+
+// [TPM2.0 1.83] 26 Miscellaneous Management Functions
+mod miscellaneous;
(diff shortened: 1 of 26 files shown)


base/src/commands.rs
```diff
@@ -14,6 +14,14 @@ pub trait TpmCommand: Marshalable {
     type RespHandles: Marshalable;
 }
 
+/// ============================================================================
```
trent-reed, line 17: These block separator comments (`// =====...`) should probably not be doc comments (e.g. don't use 3 forward slashes). They're here for the developer's sake, not the sake of the types.

You can see what happens to these comments if you run `cargo doc`, in this case the block comments are applied to the command that they are written before.

image

This is very likely not what you intended.

trent-reed, line 17: We do have a coding style against doing block comments like this though: https://github.com/tpm-rs/docs/blob/main/coding-style.md#comments

I don't mind them, and this seems like a good place to use them - the alternative is that it's hard to see at a glance which section of the specification the commands fall within. But since we have a styleguide rule against it, so we should probably follow it first.

You have a few options here:

## 1. Just don't have these block comments.

Pretty easy, just delete them.

## 2. Move each section into a submodule, and comment the submodule.

So then this file would look like this:

```rs
mod start_up;
pub use start_up::*;
mod testing;
pub use testing::*;
// ... etc. ...
```

Then in your submodules (for example, here's how `testing.rs` would look):

```rs
//! 9 Start-up [TPM2.0 1.83]

/// 9.2 _TPM_Init
pub struct InitCmd {}

/// 9.3 TPM2_Startup
#[repr(C)]
#[derive(Clone, Copy, PartialEq, Marshalable)]
pub struct StartupCmd {
    pub startup_type: TpmSu,
}
impl TpmCommand for StartupCmd {
    const CMD_CODE: TpmCc = TpmCc::Startup;
    type Handles = ();
    type RespT = ();
    type RespHandles = ();
}

/// 9.4 TPM2_Shutdown
pub struct ShutdownCmd {}
```

## 3. Seek an exception

I would not personally feel comfortable granting an exception to the rule against block comments, because I did not pen that rule. So if we want to go this route we'd have to ask that author to review.

---

Do note that technically the TPM2 specification is allowed to change the layout and numbering of the sections. So I'm not sure how useful having a bunch of section numbers in the Rust documentation is. But I also see your point in providing some structure to the command declarations here. So I'm indifferent on which path we go down, just tell me which you'd like and I'll help you out.

trent-reed:
Sorry for the delay on this - just a few comments to see what direction we should really be heading here. Let me know which path you want to head down and then I'll continue helping with the review. Thanks!

base/src/commands.rs
TRJensen: Thank you for the great input 👍 🙂 I agree, that's not what was intended.

I would probably go for removing the chapter blocks all together, and just have a single reference to TPM spec in each function request definition, which can still be used as a guide through the source code. I will try to restructure at upload again 🙂

trent-reed: You know, I say all this, but I don't really like Option (2) - so I'd avoid that one actually. Option (1) or Option (3) are okay in my book. Option (2) is a little confusing because it's not easy to tell just from the name which submodule a command would be in.

trent-reed: Thanks! Perfect, I think that works good.

TRJensen: I agree - option (2) would split all structure definitions into separate files - and some section just have a single function whereas other sections has 20 functions. As mentioned above I am currently sticking to (1), and moving the TPM reference in brackets [...] to each request. I will be pushing a suggestion momentarily 👍 🙂

oddcoder: If I am not missing anything, I think option (2) can be made sensible if we split commands into modules similar to how they are split into sections in the TPM specs.

and if there is 1-1 mapping between modules and chapters, it will make things relatively simple to upgrade each part of the implementation whenever there is spec update.

trent-reed: I'm indifferent, I'll leave it to the author.

trent-reed (approved):

oddcoder:
Hello, so from my side, I think the split is good, but please rebase to main first before merging, just to be extra safe.

base/src/commands.rs
```diff
@@ -1,6 +1,105 @@
-use crate::constants::{TpmCap, TpmCc, TpmPt, TpmSu};
+use crate::constants::TpmCc;
 use crate::Marshalable;
-use crate::{TpmiYesNo, TpmlDigest, TpmlPcrSelection, TpmsCapabilityData};
+
+/// [TPM2.0 1.83] 9 Start-up
```
oddcoder, line 4: I have 1 two little nits,


I am not sure how would this module comment be rendered, but the way to go would be doing it something like this in `startup.rs`:

``
//! [TPM2.0 1.83] 9 Start-up

use crate::commands::{Marshalable, TpmCommand};
use crate::constants::{TpmCc, TpmSu};

/// [TPM2.0 1.83] 9.2 _TPM_Init
pub struct InitCmd {}
// .. the rest of the file
```

TRJensen: Fixed

PR #111, opened 2024-10-03
Remove use of deprecated Github action with direct cargo calls

wiktor-k:
As mentioned [here](https://github.com/tpm-rs/tpm-rs/pull/108#pullrequestreview-2343220415).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,9 +19,7 @@ jobs:
           profile: minimal
           toolchain: stable
           override: true
-      - uses: actions-rs/cargo@v1
-        with:
-          command: check
+      - run: cargo check
 
   test:
     name: Test Suite
@@ -33,9 +31,7 @@ jobs:
           profile: minimal
           toolchain: stable
           override: true
-      - uses: actions-rs/cargo@v1
-        with:
-          command: test
+      - run: cargo test
 
   integration_test:
     name: Client Integration Test Suite
@@ -58,10 +54,7 @@ jobs:
           toolchain: stable
           override: true
       - run: rustup component add rustfmt
-      - uses: actions-rs/cargo@v1
-        with:
-          command: fmt
-          args: --all -- --check
+      - run: cargo fmt --all -- --check
 
   clippy:
     name: Clippy
@@ -74,10 +67,7 @@ jobs:
           toolchain: stable
           override: true
       - run: rustup component add clippy
-      - uses: actions-rs/cargo@v1
-        with:
-          command: clippy
-          args: --all-targets -- -D warnings
+      - run: cargo clippy --all-targets -- -D warnings
 
   docs:
     name: Docs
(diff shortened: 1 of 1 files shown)


oddcoder:
It look good to me, but I am not expect on CI/CD work, I would feel more comfortable having a second opinion. @trent-reed @chrisfenner what do you think about using commands directly?

In any case before merging, I would do couple of tests to make sure that CI/CD will report error if cargo commands fail th CI when they are expected to.

.github/workflows/ci.yml
```diff
@@ -19,9 +19,7 @@ jobs:
           profile: minimal
```
trent-reed, line 19: We're technically still using `actions-rs` to setup the toolchain in multiple places. The whole org is archived, I imagine that includes the toolchain repo (https://github.com/actions-rs/toolchain). Though that can be a separate PR since setup is a bit more complicated.

I assume this toolchain code wraps the rustup process, so that only matters if we're actually doing something other than stable. It's a bit annoying to repeat everywhere, so I can still see value in having an action for it if we start needing other toolchains, but currently I imagine you could even rip that out (maybe, I'm less sure of that, honestly - IDK when the containers for GitHub actions is updated to have a new rust version, so it's not clear to me that it's a no-op yet). Anyways, I am struggling to see the value of the original `actions-rs/cargo` repo.

Seems like the main thing we get from it is some support for recursing into rustup for cross-compilation, which is a feature we're not anyways using (yet): https://github.com/actions-rs/cargo/blob/master/src/main.ts

Aside from that, this is a pretty basic action that anyways ends up just calling `cargo ...`. I think we can rip it out and then if the need arises for cross-compilation we can re-evaluate what to do in the future.

trent-reed:
Should be good to not use actions here, but as @oddcoder mentioned, it's always good to check that something you expect should fail still fails. FWICT from logs, it looks like it's all still running. (for example: https://github.com/tpm-rs/tpm-rs/actions/runs/11161379200/job/31023801208?pr=111)

oddcoder:
I think it does as you can see here: https://github.com/tpm-rs/tpm-rs/pull/116

I checked how they handle things in rust, apparently ubuntu comes with rust/rustup and that is what rustorg is using, I made tiny POC and it worked as in this commit https://github.com/tpm-rs/tpm-rs/pull/116/commits/cf83995882dc9ac301963adbc11b93012c09face

@wiktor-k  Feel free to remove all the remaining dependencies on action-rs and just display the rustc version and I would merge immediately after.

Thanks in advance!

trent-reed (approved):
Approving and leaving it to the author if you want to rip out the toolchain stuff this PR or another.

wiktor-k:
Okay, there are multiple ways to solve this. The stable toolchain is the default in images prepared by Github so I just removed these lines. And the images already contain rustfmt and clippy so it's not necessary to add them too.

If you think it's a good idea to explicitly update the channel we could do something like [the CI docs say](https://github.com/rust-lang/cargo/blob/master/src/doc/src/guide/continuous-integration.md): `rustup update stable`.

I've also fixed checkout action that was in version `v2` which was super-deprecated. Fortunately `v4` is just a drop-in replacement.

Issue #118, opened 2024-10-18
Add TPM_GetRandom Command on client side.


oddcoder:
That should be implemented somewhere at [base/src/commands.rs](https://github.com/tpm-rs/tpm-rs/blob/main/base/src/commands.rs).

The implementation should be straight forward usage of `TpmCommand` similar to the rest of the file.

oddcoder:
@TRJensen would you be interested in this ?

TRJensen:
Yes - just assigned myself :-)

PR #137, opened 2024-11-06
Add structures for TPM2_GetRandom command.

TRJensen:
- Add test crude test to catch entirely random data.
- Ordered tests into sub modules matching TPM command chapters.
- Fixes https://github.com/tpm-rs/tpm-rs/issues/118

### DIFF
--- a/client/Cargo.toml
+++ b/client/Cargo.toml
@@ -12,3 +12,4 @@ tpm2-rs-marshalable = { workspace = true }
 
 [dev-dependencies]
 zerocopy = { workspace = true }
+tpm2-rs-unionify = { workspace = true }
--- a/client/tests/simulator_tests.rs
+++ b/client/tests/simulator_tests.rs
@@ -12,14 +12,16 @@
 use std::io::{Error, ErrorKind, IoSlice, Read, Result, Write};
 use std::net::TcpStream;
 use std::process::{Child, Command};
-use tpm2_rs_base::commands::{GetCapabilityCmd, StartupCmd};
-use tpm2_rs_base::constants::{TpmCap, TpmPt, TpmSu};
+use tpm2_rs_base::commands::StartupCmd;
+use tpm2_rs_base::constants::TpmSu;
 use tpm2_rs_base::errors::{TssResult, TssTcsError};
-use tpm2_rs_base::{TpmlTaggedTpmProperty, TpmsCapabilityData, TpmsTaggedProperty};
 use tpm2_rs_client::run_command;
 use tpm2_rs_client::Tpm;
 use zerocopy::big_endian::U32;
 use zerocopy::AsBytes;
+
+// Include the simulator test module.
+mod simulator_tests_registration;
 
 const SIMULATOR_IP: &str = "127.0.0.1";
 // TODO: Either pass ports or get simulator to export ports for multithreaded-use.
@@ -49,38 +51,6 @@ fn get_started_tpm() -> (TpmSim, TcpTpm) {
     };
     run_command(&startup, &mut tpm).unwrap();
     (sim_lifeline, tpm)
-}
-
-#[test]
-fn test_get_capability_manufacturer_id() {
-    let (_sim_lifeline, mut tpm) = get_started_tpm();
-
-    let mut expected = TpmlTaggedTpmProperty {
-        count: 1,
-        tpm_property: [TpmsTaggedProperty::default(); 127],
-    };
-
-    expected.tpm_property[0] = TpmsTaggedProperty {
-        property: TpmPt::Manufacturer,
-        value: 0x58595A20,
-    };
-
-    let command = GetCapabilityCmd {
-        capability: TpmCap::TPMProperties,
-        property: TpmPt::Manufacturer,
-        property_count: 1,
-    };
-
-    // We allow panic in test cases.
-    let resp = run_command(&command, &mut tpm).expect("Failed running command.");
-
-    // Extract the TpmlTaggedTpmProperty data form the response.
-    let TpmsCapabilityData::TpmProperties(received) = resp.capability_data else {
-        panic!("Unexpected variant data.")
-    };
-
-    // Confirm we received the expected response.
-    assert_eq!(received, expected);
 }
 
 // Launches the TPM simulator at the given path in a subprocess and powers it up.
--- a/client/tests/simulator_tests_registration/mod.rs
+++ b/client/tests/simulator_tests_registration/mod.rs
@@ -1,0 +1,10 @@
+//! Please see "simulator_tests.rs" for details.
+//!
+//! Add at least one test module file for each chapter in TPM command spec.
+//! If there are many tests for a single command, it is recommended to split into command submodules:
+//!
+//! <chaptername>/mod.rs
+//! <chaptername>/<commandname_1>.rs
+//! <chaptername>/<commandname_2>.rs
+pub mod capability;
+pub mod random;
--- a/base/src/commands/random.rs
+++ b/base/src/commands/random.rs
@@ -1,7 +1,27 @@
 //! [TPM2.0 1.83] 16 Random Number Generator
 
+use crate::commands::{Marshalable, TpmCommand};
+use crate::constants::TpmCc;
+use crate::Tpm2bDigest;
+
 /// [TPM2.0 1.83] 16.1 TPM2_GetRandom (Command)
-pub struct GetRandomCmd {}
+#[repr(C)]
+#[derive(Clone, Copy, PartialEq, Debug, Marshalable)]
+pub struct GetRandomCmd {
+    pub bytes_requested: u16,
+}
+impl TpmCommand for GetRandomCmd {
+    const CMD_CODE: TpmCc = TpmCc::GetRandom;
+    type Handles = ();
+    type RespT = GetRandomResp;
+    type RespHandles = ();
+}
+/// [TPM2.0 1.83] 16.1 TPM2_GetRandom (Response)
+#[repr(C)]
+#[derive(Clone, Copy, PartialEq, Debug, Marshalable)]
+pub struct GetRandomResp {
+    pub random_bytes: Tpm2bDigest,
+}
 
 /// [TPM2.0 1.83] 16.2 TPM2_StirRandom (Command)
 pub struct StirRandomCmd {}
--- a/client/tests/simulator_tests_registration/capability.rs
+++ b/client/tests/simulator_tests_registration/capability.rs
@@ -1,0 +1,39 @@
+use crate::get_started_tpm;
+use tpm2_rs_base::commands::GetCapabilityCmd;
+use tpm2_rs_base::constants::{TpmCap, TpmPt};
+use tpm2_rs_base::{TpmlTaggedTpmProperty, TpmsCapabilityData, TpmsTaggedProperty};
+use tpm2_rs_client::run_command;
+
+#[test]
+fn test_get_capability_manufacturer_id() {
+    let (_sim_lifeline, mut tpm) = get_started_tpm();
+
+    let mut expected = TpmlTaggedTpmProperty {
+        count: 1,
+        tpm_property: [TpmsTaggedProperty::default(); 127],
+    };
+
+    expected.tpm_property[0] = TpmsTaggedProperty {
+        property: TpmPt::Manufacturer,
+        value: 0x58595A20,
+    };
+
+    let command = GetCapabilityCmd {
+        capability: TpmCap::TPMProperties,
+        property: TpmPt::Manufacturer,
+        property_count: 1,
+    };
+
+    // We allow panic in test cases.
+    let resp = run_command(&command, &mut tpm).expect("Failed running command.");
+
+    // Extract the TpmlTaggedTpmProperty data form the response.
+    let TpmsCapabilityData::TpmProperties(received) = resp.capability_data else {
+        panic!("Unexpected variant data.")
+    };
+
+    assert_eq!(
+        received, expected,
+        "Received did not match the expected response."
+    );
+}
--- a/client/tests/simulator_tests_registration/random.rs
+++ b/client/tests/simulator_tests_registration/random.rs
@@ -1,0 +1,95 @@
+use crate::get_started_tpm;
+use tpm2_rs_base::commands::GetRandomCmd;
+use tpm2_rs_base::constants::TPM2_SHA256_DIGEST_SIZE;
+use tpm2_rs_base::TpmtHa;
+use tpm2_rs_client::run_command;
+use tpm2_rs_unionify::UnionSize;
+
+#[test]
+fn test_get_random_duplicate_value_trap() {
+    let (_sim_lifeline, mut tpm) = get_started_tpm();
+
+    let command = GetRandomCmd {
+        bytes_requested: TPM2_SHA256_DIGEST_SIZE as u16,
+    };
+
+    let resp = run_command(&command, &mut tpm).expect("Failed running command.");
+
+    // Lets pull out the actual data as a slice for convenience
+    let random_slice = &resp.random_bytes.as_ref();
+
+    assert_eq!(
+        random_slice.len(),
+        TPM2_SHA256_DIGEST_SIZE as usize,
+        "We should have received exactly size of SHA256 bytes, but got {}.",
+        random_slice.len()
+    );
+
+    // Duplicate value trap, to catch any same value sequences.
+    let same_twice_occurrences = random_slice
+        .iter()
+        .zip(random_slice.iter().skip(1))
+        .filter(|(prev, curr)| prev == curr)
+        .count();
+
+    assert!(
+        same_twice_occurrences < (random_slice.len() / 2).into(),
+        "More than 50% of the values equals previous value: {random_slice:?}"
+    );
+}
+
+#[test]
+fn test_get_random_large_sizes() {
+    let (_sim_lifeline, mut tpm) = get_started_tpm();
+    let mut detected_max_size = 0;
+
+    // The first value is used to detect the servers max digest size.
+    // The second value is used to confirm that server is still providing that size.
+    for i in [0xFFF, 0xFFFF] {
+        let command = GetRandomCmd { bytes_requested: i };
+        let resp = run_command(&command, &mut tpm).expect("Failed running command.");
+
+        // Lets pull out the actual slice size for convenience
+        let random_slice_len = resp.random_bytes.as_ref().len();
+
+        if detected_max_size == 0 {
+            // Detect the max size used by the server.
+            assert!(
+                TpmtHa::UNION_SIZE as usize >= random_slice_len,
+                "We received more random data, than client implementation supports {random_slice_len} > {}.",
+                TpmtHa::UNION_SIZE
+            );
+
+            assert!(
+                TPM2_SHA256_DIGEST_SIZE as usize <= random_slice_len,
+                "We should have received at least size of SHA256 bytes, but got {random_slice_len}."
+            );
+
+            detected_max_size = random_slice_len;
+        } else {
+            assert_eq!(
+                detected_max_size, random_slice_len,
+                "We should have received max size {detected_max_size}, but got {random_slice_len}.",
+            );
+        }
+    }
+}
+
+#[test]
+fn test_get_random_small_sizes() {
+    let (_sim_lifeline, mut tpm) = get_started_tpm();
+
+    for i in 0..1 {
+        let command = GetRandomCmd { bytes_requested: i };
+
+        let resp = run_command(&command, &mut tpm).expect("Failed running command.");
+
+        // Lets pull out the actual slice size for convenience
+        let random_slice_len = resp.random_bytes.as_ref().len();
+
+        assert_eq!(
+            random_slice_len, i as usize,
+            "We should have received {i}, but got {random_slice_len} bytes."
+        );
+    }
+}


client/tests/simulator_tests.rs
```diff
@@ -83,6 +83,63 @@ fn test_get_capability_manufacturer_id() {
...
 
+#[test]
+fn test_get_random() {
+    const REQUEST_SIZE: u16 = 32;
+    const MIN_RECEIVED: u16 = 32;
+
+    let (_sim_lifeline, mut tpm) = get_started_tpm();
+
+    let command = GetRandomCmd {
+        bytes_requested: REQUEST_SIZE,
```
trent-reed, line 94: Personally I'm not sure I see the need for a constant for this, seems to make the test less-readable IMO.

client/tests/simulator_tests.rs
```diff
@@ -83,6 +83,63 @@ fn test_get_capability_manufacturer_id() {
...
+    };
+
+    // We allow panic in test cases.
+    let resp = run_command(&command, &mut tpm).expect("Failed running command.");
+
+    // Validate that retrieved data contains at least 32 values (SHA256 digest),
+    // and is less that requested amount.
+    assert!(resp.random_bytes.size >= MIN_RECEIVED);
```
trent-reed, line 102: Correct me if I'm wrong, but it should only ever be less-than the requested size (in the case where the requested bytes are greater than the size of a `Tpm2bDigest`)?

client/tests/simulator_tests.rs
```diff
@@ -83,6 +83,63 @@ fn test_get_capability_manufacturer_id() {
...
+    assert!(resp.random_bytes.size <= REQUEST_SIZE);
```
trent-reed, line 103: I wonder if we should assert instead <= size of the `Tpm2bDigest`?

client/tests/simulator_tests.rs
```diff
@@ -83,6 +83,63 @@ fn test_get_capability_manufacturer_id() {
...
+
+    // Lets pull out the actual data as a slice for convenience
+    let random_slice = &resp.random_bytes.buffer[0..resp.random_bytes.size as usize];
```
trent-reed, line 106: You don't need to state the first 0 in `[0..*]`, if you leave it off it's implied.

client/tests/simulator_tests.rs
```diff
@@ -83,6 +83,63 @@ fn test_get_capability_manufacturer_id() {
     assert_eq!(received, expected);
 }
 
+#[test]
+fn test_get_random() {
```
trent-reed, line 87: Probably worth having a test that checks that if you ask for a size larger than `Tpm2bDigest` that it cuts back the response properly and doesn't fail or panic.

client/tests/simulator_tests.rs
```diff
@@ -83,6 +83,63 @@ fn test_get_capability_manufacturer_id() {
     assert_eq!(received, expected);
 }
 
+#[test]
```
trent-reed, line 86: Test structure idea: I can see this is going to get unweildy as more commands are implemented. WDYT of each command having a separate module and then you can implement tests specific to commands in those submodules?

Then you can name the tests more specifically what they test and it'll look nicer in test output.

So for instance, `simulator_tests::get_random::trivial_repeat_check` would be this test, and `simulator_tests::get_random::clamps_to_tpm2_digest_size` would be the other test I've asked for below.

client/tests/simulator_tests.rs
```diff
@@ -83,6 +83,63 @@ fn test_get_capability_manufacturer_id() {
...
+        let mut i = 0;
+        for i in 0..random_slice.len() as usize {
+            // Calculate the circular distance modulo 256 going upwards from first value.
+            // This will give an even probability distribution of values [0 ; 255].
+            let diff: u16 = (256 as u16 + random_slice[i] as u16 - last_value as u16) % 256;
+            count[(diff >> 4) as usize] += 1;
+            last_value = random_slice[i];
+        }
+
+        for i in 0..count.len() as usize {
```
trent-reed, line 125: Can be expressed more succinctly:

```rs
let zeros = count.iter().filter(|v| v == 0).count();
```

client/tests/simulator_tests.rs
```diff
@@ -83,6 +83,63 @@ fn test_get_capability_manufacturer_id() {
...
+    // Print for test inspection.
+    println!("Got random sequence: {:?}", random_slice);
+
+    // Crude randomness test. Confirm we did not get identical data sequences.
+    {
+        let mut zeros: u16 = 0;
+        let mut count: [u16; 16] = [0; 16]; // 16 slots.
+        let mut last_value = random_slice[random_slice.len() - 1];
+        let mut i = 0;
+        for i in 0..random_slice.len() as usize {
```
trent-reed, line 117: It might be slightly nicer to use iterators to get your `prev` and `next` instead of relying on indexing the slice:

```rs
for (prev, curr) in random_slice.iter().zip(random_slice.iter().cycle().skip(1)) {
    // Your calculation.
}
```

Though I'm not sure if my brain is just broken from doing too much work with iterators. So take this as an optional. Note that it doesn't technically iterate in the exact same order, but I think it accomplishes the same thing, as it iterates over all pairs of `(prev, curr)` (the difference is the pair `([len()-1],[0])` happens at the end in the iterator construction above).

client/tests/simulator_tests.rs
```diff
@@ -83,6 +83,63 @@ fn test_get_capability_manufacturer_id() {
...
+
+    // Print for test inspection.
+    println!("Got random sequence: {:?}", random_slice);
+
+    // Crude randomness test. Confirm we did not get identical data sequences.
+    {
+        let mut zeros: u16 = 0;
+        let mut count: [u16; 16] = [0; 16]; // 16 slots.
+        let mut last_value = random_slice[random_slice.len() - 1];
+        let mut i = 0;
```
trent-reed, line 116: I wouldn't make this local, not needed.

client/tests/simulator_tests.rs
```diff
@@ -83,6 +83,63 @@ fn test_get_capability_manufacturer_id() {
...
+    let random_slice = &resp.random_bytes.buffer[0..resp.random_bytes.size as usize];
+
+    // Print for test inspection.
+    println!("Got random sequence: {:?}", random_slice);
+
+    // Crude randomness test. Confirm we did not get identical data sequences.
+    {
+        let mut zeros: u16 = 0;
+        let mut count: [u16; 16] = [0; 16]; // 16 slots.
+        let mut last_value = random_slice[random_slice.len() - 1];
```
trent-reed, line 115: `last_value` not needed if you use iterators as detailed below (optional).

client/tests/simulator_tests.rs
```diff
@@ -83,6 +83,63 @@ fn test_get_capability_manufacturer_id() {
...
+        let mut i = 0;
+        for i in 0..random_slice.len() as usize {
+            // Calculate the circular distance modulo 256 going upwards from first value.
+            // This will give an even probability distribution of values [0 ; 255].
+            let diff: u16 = (256 as u16 + random_slice[i] as u16 - last_value as u16) % 256;
```
trent-reed, line 120: I assume this `256` is to avoid underflow, but personally I'd just embrace it and use `u8` for your logic here: https://doc.rust-lang.org/std/primitive.u8.html#method.wrapping_sub

client/tests/simulator_tests.rs
```diff
@@ -83,6 +83,63 @@ fn test_get_capability_manufacturer_id() {
...
+    // Validate that retrieved data contains at least 32 values (SHA256 digest),
+    // and is less that requested amount.
+    assert!(resp.random_bytes.size >= MIN_RECEIVED);
+    assert!(resp.random_bytes.size <= REQUEST_SIZE);
+
+    // Lets pull out the actual data as a slice for convenience
+    let random_slice = &resp.random_bytes.buffer[0..resp.random_bytes.size as usize];
+
+    // Print for test inspection.
+    println!("Got random sequence: {:?}", random_slice);
```
trent-reed, line 109: Can be written as `println!("Got random sequence: {random_slice:?}");`

client/tests/simulator_tests.rs
```diff
@@ -83,6 +83,63 @@ fn test_get_capability_manufacturer_id() {
...
+            count[(diff >> 4) as usize] += 1;
+            last_value = random_slice[i];
+        }
+
+        for i in 0..count.len() as usize {
+            if count[i] == 0 {
+                zeros += 1;
+            }
+        }
+        println!(
```
trent-reed, line 130: Can be written as: 

```rs
        println!(
            "Filled circular distance of RND bytes into {count} slots: {}",
            count.len()
        );
```

client/tests/simulator_tests.rs
```diff
@@ -83,6 +83,63 @@ fn test_get_capability_manufacturer_id() {
...
+            "Filled circular distance of RND bytes into {} slots: {:?}",
+            count.len(),
+            count
+        );
+        println!(
```
trent-reed, line 135: Can be written as:

```rs
        println!(
            "After drawing 32 values, we found {zeros} empty slots"
        );
```

Not sure we need to say how many we allow, the assertion will say that for us, you can add additional text to `assert` by the way, if you'd like to provide the added easy-to-read English context there: https://doc.rust-lang.org/std/macro.assert.html

trent-reed (commented):
Reviewed, thanks! Take what makes sense and push back on what doesn't.

client/tests/simulator_tests.rs
oddcoder, line 87: I have one comment here, I think this test is overkill, to my understanding we do basic statistical test for randomness. It can be simplified to making sure that the result array is not all zeros.

Also I would like to see more tests like getting zero bytes of randomness, 1 byte of randomness, and trying to get more bytes than the limits.

client/tests/simulator_tests.rs
TRJensen: Cleaned up

client/tests/simulator_tests.rs
TRJensen: Fixed 👍

client/tests/simulator_tests.rs
TRJensen: Fixed 👍

client/tests/simulator_tests.rs
TRJensen: Fixed 👍

client/tests/simulator_tests.rs
TRJensen: I have simplified the test - so construct no longer used 👍

client/tests/simulator_tests.rs
TRJensen: I have simplified the test - so construct no longer used 👍

client/tests/simulator_tests.rs
TRJensen: Changed to test for simple repeated values test - should catch the typical memory dumps like 00 00 00 01 00 00 00 00 or typical memset patterns different from 0.

client/Cargo.toml
```diff
@@ -9,6 +9,8 @@ tpm-simulator-tests = []
 [dependencies]
 tpm2-rs-base = { workspace = true }
 tpm2-rs-marshalable = { workspace = true }
+tpm2-rs-unionify = { workspace = true }
+tracing = { workspace = true }
```
trent-reed, line 13: :(

Hard disagree.

If you needed this for testing, I would have added it to dev-dependencies.

trent-reed: Oh, I thought you already merged this, but it's actually still in review - my bad.

Anyways, I don't think we need this, personally. I don't see what this really adds that we don't get well-enough with just `println` in test situations, but I'd be okay adding it as long as you only add it to `dev-dependencies`

client/Cargo.toml
```diff
@@ -9,6 +9,8 @@ tpm-simulator-tests = []
 [dependencies]
 tpm2-rs-base = { workspace = true }
 tpm2-rs-marshalable = { workspace = true }
+tpm2-rs-unionify = { workspace = true }
```
trent-reed, line 12: Any particular reason the client is adding this? Probably for sizing in tests. In that case please move to dev-dependencies.

client/tests/simulator_tests.rs
```diff
@@ -83,6 +85,66 @@ fn test_get_capability_manufacturer_id() {
...
+    };
+
+    // We allow panic in test cases.
+    let resp = run_command(&command, &mut tpm).expect("Failed running command.");
+
+    // Lets pull out the actual data as a slice for convenience
+    let random_slice = &resp.random_bytes.buffer[..resp.random_bytes.size as usize];
+
+    // We should have received less than what the client implementation supports.
+    assert!(random_slice.len() <= TpmtHa::UNION_SIZE);
```
trent-reed, line 103: Optional: Instead of having the comments here, you could mix them into the assert so that they print on failure.

e.g. `assert!(..., "We expected to have received no more than the maximum supported digest size")`

client/tests/simulator_tests.rs
```diff
@@ -83,6 +85,66 @@ fn test_get_capability_manufacturer_id() {
...
+
+    // We should have received minimum size of SHA256 (minimum mandatory digest size).
+    assert!(random_slice.len() >= TPM2_SHA256_DIGEST_SIZE as usize);
+
+    // Print for test inspection.
+    trace!("Got random sequence: {:?}", random_slice);
```
trent-reed, line 109: Maybe just me, but I don't think I'd take a crate dependency for debug prints in tests. You only see the prints when the test fails anyways, and I think that's a desirable trait, myself. Why are we adding this `trace` stuff?

client/tests/simulator_tests.rs
```diff
@@ -83,6 +85,66 @@ fn test_get_capability_manufacturer_id() {
...
+
+    // Duplicate value trap, to catch any same value sequences.
+    {
+        let mut same_twice: u16 = 0;
+        for i in 1..random_slice.len() as usize {
```
trent-reed, line 114: Optional: Could be done with iterators; I find this more readable though not all may:

```rs
let same_twice_occurrences = random_slice.iter()
    .zip(random_slice.iter().skip(1))
    .filter(|(prev, next)| prev == next)
    .count();
```

---

Ideally Rust would provide a window function, which is what you really want here, but I don't know that I want to pull in itertools for a single function. Seems likely to someday be in the std so I guess we can just build the window manually for now using `zip`.

client/tests/simulator_tests.rs
```diff
@@ -83,6 +85,66 @@ fn test_get_capability_manufacturer_id() {
...
+        };
+
+        // We allow panic in test cases.
+        let resp = run_command(&command, &mut tpm).expect("Failed running command.");
+
+        // Lets pull out the actual data as a slice for convenience
+        let random_slice = &resp.random_bytes.buffer[..resp.random_bytes.size as usize];
+
+        // We should have received zero bytes.
+        assert!(random_slice.len() == i);
```
trent-reed, line 144: Probably better to `assert_eq!` instead.

trent-reed: Also, you probably don't need to form the slice, you could probably just inspect `resp.random_bytes.size` for this test.

client/tests/simulator_tests.rs
```diff
@@ -83,6 +85,66 @@ fn test_get_capability_manufacturer_id() {
...
+        );
+        assert!(same_twice <= random_slice.len() as u16 / 2); // Same value repeated in half of the values unlikely with > 32 values.
+    }
+}
+
+#[test]
+fn test_get_random_small_sizes() {
+    let (_sim_lifeline, mut tpm) = get_started_tpm();
+
+    for i in 0..4 as usize {
```
trent-reed, line 132: Why up to `4`? Just to test the lower-level functionality from `u32` vs `u64` that might be treated specially (ala @oddcoder's ongoing PR)?

In that case maybe worth going up to size of `u64` or `u128` to set some clear obvious boundary defined by a type.

trent-reed, line 132: Unneeded `as usize` cast, I think

trent-reed (commented):
Re-reviewed, thanks!

client/tests/simulator_tests.rs
TRJensen: I love this Rust construct 👍 I played a little around with it 🙂 The zip will only provide a value pair iterator corresponding to the shorter of the two iterators. Cycle() loops the one iterator, so it infinitely wraps around. And skip offsets the iterator 1 element in this case. In the simplified crude test, I just need adjacent value pairs, so I adopted this one and skipped the cycle 👍

client/Cargo.toml
TRJensen: I have moved all texts into the assert line, so that we only get the print in case of a failed test 👍 🙂

client/Cargo.toml
TRJensen: Moved 👍 🙂

client/tests/simulator_tests.rs
TRJensen: Fixed.

client/tests/simulator_tests.rs
```diff
@@ -83,6 +85,66 @@ fn test_get_capability_manufacturer_id() {
...
+    let resp = run_command(&command, &mut tpm).expect("Failed running command.");
+
+    // Lets pull out the actual data as a slice for convenience
+    let random_slice = &resp.random_bytes.buffer[..resp.random_bytes.size as usize];
+
+    // We should have received less than what the client implementation supports.
+    assert!(random_slice.len() <= TpmtHa::UNION_SIZE);
+
+    // We should have received minimum size of SHA256 (minimum mandatory digest size).
+    assert!(random_slice.len() >= TPM2_SHA256_DIGEST_SIZE as usize);
```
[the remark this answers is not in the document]
TRJensen: Fixed

client/tests/simulator_tests.rs
TRJensen: Original idea was to add tracing so it can be disabled/enabled at will - but adding messages to the assert messages is fine here. Fixed 👍 🙂

client/tests/simulator_tests.rs
TRJensen: Ah - another great Rust construct 🙂 I am still thinking too much in the "C" programming strategies. But here the Rust compiler will optimize (i hope) by it self. I have adopted this strategy - thanks again for treat tips 🙂 👍

client/tests/simulator_tests.rs
```diff
@@ -83,6 +85,66 @@ fn test_get_capability_manufacturer_id() {
...
+            if random_slice[i - 1] == random_slice[i] {
+                same_twice += 1;
+            }
+        }
+
+        trace!(
+            "After {} values, we got duplicate value {same_twice} times.",
+            random_slice.len()
+        );
+        assert!(same_twice <= random_slice.len() as u16 / 2); // Same value repeated in half of the values unlikely with > 32 values.
```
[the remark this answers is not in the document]
TRJensen: Fixed

client/tests/simulator_tests.rs
```diff
@@ -83,6 +85,66 @@ fn test_get_capability_manufacturer_id() {
...
+            bytes_requested: i as u16,
+        };
+
+        // We allow panic in test cases.
+        let resp = run_command(&command, &mut tpm).expect("Failed running command.");
+
+        // Lets pull out the actual data as a slice for convenience
+        let random_slice = &resp.random_bytes.buffer[..resp.random_bytes.size as usize];
+
+        // We should have received zero bytes.
```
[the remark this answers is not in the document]
TRJensen: Removed

client/tests/simulator_tests.rs
TRJensen: Fixed

client/tests/simulator_tests.rs
TRJensen: Fixed

client/tests/simulator_tests.rs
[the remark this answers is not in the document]
TRJensen: Changed to zip() skip() filter() count() construct 🙂 👍 

Fixed

client/tests/simulator_tests.rs
TRJensen: I think it is a great idea. We already moved all command structures into sub folders. And these tests may explode in future. I will move it once we are done reviewing the current test cases 👍 🙂 Or we can create a separate task for this cleanup, no to have too many changes in one go.

client/tests/simulator_tests.rs
TRJensen: Fixed (used similar in new construct).

client/tests/simulator_tests.rs
TRJensen: Fixed (same idea brought forward to latest changes).

client/tests/simulator_tests.rs
TRJensen: I have changed to test just 0 and 1. Zero requests seems to be allowed for some reason. And 1 is the minimal case.
The number of requested bytes is (by TPM spec and our request struct definition) limited to u16. So I modified the max test to handle 2 large requests. (0xFFF and 0xFFFF).

base/src/lib.rs
```diff
@@ -584,49 +584,49 @@ pub type Tpm2bOperand = Tpm2bDigest;
 #[repr(C)]
 #[derive(Clone, Copy, PartialEq, Debug)]
 pub struct Tpm2bData {
-    size: u16,
+    pub size: u16,
```
oddcoder, line 587: I do not think tpm2b family of types should expose size and buffer content. It might be better to expose that functionality over 3 API.

```
impl AsRef<[u8]> for Tpm2b*

impl AsMut<[u8]> for Tpm2b*
```

client/tests/simulator/capability.rs
```diff
@@ -0,0 +1,42 @@
+#![cfg(feature = "tpm-simulator-tests")]
```
trent-reed, line 1: Unneeded, there's already a conditional compile in the parent module.

client/tests/simulator/mod.rs
```diff
@@ -0,0 +1,9 @@
+#![cfg(feature = "tpm-simulator-tests")]
```
trent-reed, line 1: Unneeded, there's already a conditional compile in the parent module.

client/tests/simulator/random.rs
```diff
@@ -0,0 +1,98 @@
+#![cfg(feature = "tpm-simulator-tests")]
```
trent-reed, line 1: Unneeded, there's already a conditional compile in the parent module.

client/tests/simulator/mod.rs
```diff
@@ -0,0 +1,9 @@
...
+
+// Please see "simulator_tests.rs" for details.
```
trent-reed, line 3: Yeah, this is where we get into something I dislike about current rust language - the `tests` directory is kind of magical, in that it does something that we wouldn't expect with module resolution...

What I wish we could do is:

*   client/tests/simulator_tests/main.rs
*   client/tests/simulator_tests/submodule_a.rs
*   client/tests/simulator_tests/submodule_b.rs

But you can't do that. Instead you have to create a helper module and import that (like you're doing here). This is fine, and allows us to structure and organize our code, but the naming feels a little odd. Naming it `simulator` sounds like it's a module for providing the simulator, not for providing tests relating to the simulator.

---

So there's two options here:

1.  Turn `simulator` into a shared module that provides code for how to interact with the simulator, and then each test is it's own separate integration test (that's a bit annoying, I think; though we can reconsider if we want this in a future PR, we probably shouldn't change to this for now).
2.  Name the module explicitly to suggest it's about organizing test code for `simulator_tests`.

This might be bikeshedding a bit, so feel free to ignore, but WDYT of naming this module `simulator_tests_registration`, to make it clear it's a helper module that is responsible for organizing and registering tests related to `simulator_tests`.

So the final layout would look like this:

*   client/tests/simulator_tests_registration/mod.rs
*   client/tests/simulator_tests_registration/random.rs
*   client/tests/simulator_tests_registration/capability.rs
*   client/tests/simulator_tests.rs

Then in `simulator_tests.rs` we include that module:

```rs
mod simulator_tests_registration;
```

---

This is basically exactly what you're doing here, except that I'm suggesting we rename the `simulator` module to `simulator_tests_registration` to imply a more specific relation to `simulator_tests`.

client/tests/simulator/capability.rs
```diff
@@ -0,0 +1,42 @@
...
+
+// This module is added to "mod.rs" to be included in the test. See "mod.rs" for details.
```
trent-reed, line 3: Not sure this is a necessary comment; the comment about test structure and organization in the sibling `mod.rs` is probably sufficient IMHO.

client/tests/simulator/random.rs
```diff
@@ -0,0 +1,98 @@
...
+
+// This module is added to "mod.rs" to be included in the test. See "mod.rs" for details.
```
trent-reed, line 3: Not sure this is a necessary comment; the comment about test structure and organization in the sibling `mod.rs` is probably sufficient IMHO.

client/tests/simulator/mod.rs
```diff
@@ -0,0 +1,9 @@
...
+//
+// Add at least one test module file for each chapter in TPM command spec.
+// If there are many tests for a single command, it is recommended to split into command specific files.
+// For example: <chaptername>_<commandname>.rs
```
trent-reed, line 7: My recommendation would be that they create further submodules. E.g.

*   `<chaptername>/mod.rs`
*   `<chaptername>/<commandname_1>.rs`
*   `<chaptername>/<commandname_2>.rs`

trent-reed: The only place where submodule resolution is weird is directly under the `tests` directory, now that we're under a normal module things should work in a pretty expected manner.

client/tests/simulator_tests.rs
trent-reed: I left a separate comment here around the oddities of `tests` module resolution. I kind of forgot about that, so I hope it didn't cause too much headaches... Anyways, I think it's worth the headaches, less we eventually have a several thousand-line `simulator_tests.rs` file which will be impossible to maintain...

Thanks for restructuring!

client/tests/simulator_tests.rs
```diff
@@ -83,6 +83,63 @@ fn test_get_capability_manufacturer_id() {
     assert_eq!(received, expected);
 }
 
+#[test]
+fn test_get_random() {
+    const REQUEST_SIZE: u16 = 32;
+    const MIN_RECEIVED: u16 = 32;
+
+    let (_sim_lifeline, mut tpm) = get_started_tpm();
```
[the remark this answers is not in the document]
trent-reed: I opened an issue for this: https://github.com/tpm-rs/tpm-rs/issues/139

client/Cargo.toml
trent-reed: Thanks! I think this is the way forward with our tests (adding details to the assertions), so hopefully that's all we need. :)

client/tests/simulator_tests.rs
trent-reed: It should! Iterators are one of those types that attempt to inline and be written as-if you had written them by hand.

In general, you should feel empowered to use the helper functions of `Option`, `Result`, and `Iterator` as-if you were writing their inner code out by hand; this is very different from what we usually think in C and C++.

My experience has been the compiler is pretty good about this stuff. Most of the time producing something as if you had written the trickier for-loop out yourself. So I think we should trust the compiler for these constructs until we have a situation where we need to be *very certain* about the efficiency of something (and tests I'm not too worried about efficiency, so I'd opt for whatever is most readable here regardless of the cost, personally).

Anyways, this is one of the things that Rust means when it promises these "zero-cost abstractions".

trent-reed:
Okay, next pass done - thanks!

client/tests/simulator_tests.rs
```diff
@@ -83,6 +83,63 @@ fn test_get_capability_manufacturer_id() {
     assert_eq!(received, expected);
 }
 
+#[test]
+fn test_get_random() {
+    const REQUEST_SIZE: u16 = 32;
+    const MIN_RECEIVED: u16 = 32;
+
```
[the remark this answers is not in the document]
TRJensen: Fixed

client/tests/simulator_tests.rs
TRJensen: Thanks for the great clarification 👍 🙂

client/tests/simulator/capability.rs
TRJensen: Fixed

client/tests/simulator/mod.rs
TRJensen: Fixed

client/tests/simulator/random.rs
TRJensen: Fixed

client/tests/simulator/mod.rs
TRJensen: Super suggestion - I will adopt 🙂 I don't mind the "bikeshedding" (a word I had to look up, due to English being my 2nd language 😄 ).  It is important to get these details in place from the start, to avoid larger cleanups in the future 👍

client/tests/simulator/mod.rs
TRJensen: I have updated the text recommendation. Fixed.

TRJensen:
@trent-reed @oddcoder Some more adjustments and restructuring based on comments. Please have a final look 👍 🙂

client/tests/simulator_tests_registration/mod.rs
```diff
@@ -0,0 +1,9 @@
+// Please see "simulator_tests.rs" for details.
```
oddcoder, line 1: Preferably use doc comments and not those styles

```
//! Please see ...
```

oddcoder (approved):
So from my side, I just have one minor suggestion, but aside from that everything else look good.

client/tests/simulator_tests_registration/mod.rs
TRJensen: Fixed

PR #148, opened 2024-12-11
Remove a newline between doc comment and code.

trent-reed:
This was causing trouble for CI clippy checks (source: https://github.com/tpm-rs/tpm-rs/actions/runs/12285661194), it's been that way for a while, so I think we just missed it when enabling the clippy checks in CI.

I checked this using `cargo clippy --all-targets -- -D warnings` manually and got a good result.

### DIFF
--- a/client/src/sessions/password.rs
+++ b/client/src/sessions/password.rs
@@ -39,7 +39,6 @@ impl PasswordSession {
     ///     PasswordSession::new(&bad_password).err().unwrap(),
     ///     TpmRcError::Size
     /// );
-
     /// ```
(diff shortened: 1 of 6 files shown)


trent-reed:
Uh-oh, looks like there are other failures in clippy as well - weird. I wonder why these weren't noticed when we first enabled clippy support?

oddcoder:
Newer rust release, with every new release they stabilize more warnings, anyways one it passes just squash and mere please.

oddcoder (approved):

trent-reed:
> Newer rust release, with every new release they stabilize more warnings, anyways one it passes just squash and mere please.

Yup that's it, after a `rustup` I can repro the errors on my box. I'll fix and update.

PR #150, opened 2025-01-10
Improve Tpm2b* and Tpm2bSimple traits.

### LINKED ISSUE #114: Improve status of Tpm2b* and Tpm2bSimple trait
We have a macro called `impl_try_marshalable_tpm2b_simple` which may be unnessary and not idiomatic for the case it is handling.

The macro is deriving three traits:
1- `Default`: Based on my intuition, a `#[derive(Default)]` would actually suffice.
2- `Marshalable`: which is the responsibility of `marshalable` + `marshalable-derive`
3- `Tpm2bSimple`: In the implementation of that macro we are doing the following:
```rust
impl Tpm2bSimple for $T {
    const MAX_BUFFER_SIZE: usize = size_of::<$T>() - size_of::<u16>();
    // more stuff
}
``` 
But the way `MAX_BUFFER_SIZE` is defined would be undefined behavior under `#repr(rust)`. It makes implicit assumption that `#[repr(C)]` is there, but it will silently go to UB once that `repr` is removed.

Ideally we can do the following:
1- **Auto derive Default**: and test at least for one of the cases that we get the values we expect.
2- **Handle Marshalable properly**: One idea is to have custom logic for `Marshalable` to handle `Tpm2b*`. It can look as follows:
```rust
#[repr(C)]
#[derive(Clone, Copy, PartialEq, Debug, Marshalable)]
#[Marshalable(Tpm2b)] // note this line
pub struct Tpm2bCreationData {
    size: u16,
    pub creation_data: [u8; size_of::<TpmsCreationData>()],
}
```
and then it would generate appropriate implementation for Marshalable similar to what is defined .
3- **Handle Tpm2bSimple Properly** : Possibly moving the trait to `marshalable` crate? and having a proc macro that does one of two things:
1- verify that the struct in question is `#[repr(c)]` and fail with error message otherwise.
2- figures another way to derive `MAX_BUFFER_SIZE`

TRJensen:
- Replace "impl_try_marshalable_tpm2b_simple" with derive attribute #[marshalable(tpm2b_simple)].
- Replace "impl_try_marshalable_tpm2b_struct" with #[derive(Tpm2bStruct)].
- Fixes https://github.com/tpm-rs/tpm-rs/issues/114.

### DIFF
--- a/base/src/lib.rs
+++ b/base/src/lib.rs
@@ -854,9 +854,8 @@ pub struct TpmsSensitiveCreate {
 }
 
 #[repr(C)]
-#[derive(Clone, Copy, PartialEq, Debug, Marshalable)]
+#[derive(Clone, Copy, PartialEq, Debug, Marshalable, Tpm2bStruct)]
 #[marshalable(tpm2b_simple)]
-#[marshalable(tpm2b_struct)]
 pub struct Tpm2bSensitiveCreate {
     size: u16,
     sensitive: [u8; size_of::<TpmsSensitiveCreate>()],
@@ -1166,9 +1165,8 @@ impl Marshalable for TpmtPublic {
 }
 
 #[repr(C)]
-#[derive(Clone, Copy, PartialEq, Debug, Marshalable)]
+#[derive(Clone, Copy, PartialEq, Debug, Marshalable, Tpm2bStruct)]
 #[marshalable(tpm2b_simple)]
-#[marshalable(tpm2b_struct)]
 pub struct Tpm2bPublic {
     size: u16,
     public_area: [u8; size_of::<TpmtPublic>()],
@@ -1457,9 +1455,8 @@ pub struct TpmsCreationData {
 }
 
 #[repr(C)]
-#[derive(Clone, Copy, PartialEq, Debug, Marshalable)]
+#[derive(Clone, Copy, PartialEq, Debug, Marshalable, Tpm2bStruct)]
 #[marshalable(tpm2b_simple)]
-#[marshalable(tpm2b_struct)]
 pub struct Tpm2bCreationData {
     size: u16,
     creation_data: [u8; size_of::<TpmsCreationData>()],
--- a/base/src/lib.rs
+++ b/base/src/lib.rs
@@ -589,7 +589,8 @@ enum TpmuName {
 }
 
 #[repr(C)]
-#[derive(Clone, Copy, PartialEq, Debug)]
+#[derive(Clone, Copy, PartialEq, Debug, Marshalable)]
+#[marshalable(tpm2b_simple)]
 pub struct Tpm2bDigest {
     size: u16,
     buffer: [u8; TpmtHa::UNION_SIZE],
@@ -599,42 +600,48 @@ pub type Tpm2bNonce = Tpm2bDigest;
 pub type Tpm2bOperand = Tpm2bDigest;
 
 #[repr(C)]
-#[derive(Clone, Copy, PartialEq, Debug)]
+#[derive(Clone, Copy, PartialEq, Debug, Marshalable)]
+#[marshalable(tpm2b_simple)]
 pub struct Tpm2bData {
     size: u16,
     buffer: [u8; TpmtHa::UNION_SIZE],
 }
 
 #[repr(C)]
-#[derive(Clone, Copy, PartialEq, Debug)]
+#[derive(Clone, Copy, PartialEq, Debug, Marshalable)]
+#[marshalable(tpm2b_simple)]
 pub struct Tpm2bEvent {
     size: u16,
     buffer: [u8; 1024],
 }
 
 #[repr(C)]
-#[derive(Clone, Copy, PartialEq, Debug)]
+#[derive(Clone, Copy, PartialEq, Debug, Marshalable)]
+#[marshalable(tpm2b_simple)]
 pub struct Tpm2bMaxBuffer {
     size: u16,
     buffer: [u8; TPM2_MAX_DIGEST_BUFFER as usize],
 }
 
 #[repr(C)]
-#[derive(Clone, Copy, PartialEq, Debug)]
+#[derive(Clone, Copy, PartialEq, Debug, Marshalable)]
+#[marshalable(tpm2b_simple)]
 pub struct Tpm2bMaxNvBuffer {
     size: u16,
     buffer: [u8; TPM2_MAX_NV_BUFFER_SIZE as usize],
 }
 
 #[repr(C)]
-#[derive(Clone, Copy, PartialEq, Debug)]
+#[derive(Clone, Copy, PartialEq, Debug, Marshalable)]
+#[marshalable(tpm2b_simple)]
 pub struct Tpm2bIv {
     size: u16,
     buffer: [u8; TPM2_MAX_SYM_BLOCK_SIZE as usize],
 }
 
 #[repr(C)]
-#[derive(Clone, Copy, PartialEq, Debug)]
+#[derive(Clone, Copy, PartialEq, Debug, Marshalable)]
+#[marshalable(tpm2b_simple)]
 pub struct Tpm2bName {
     size: u16,
     name: [u8; TpmuName::UNION_SIZE],
@@ -786,21 +793,24 @@ impl Marshalable for TpmsAttest {
 }
 
 #[repr(C)]
-#[derive(Clone, Copy, PartialEq, Debug)]
+#[derive(Clone, Copy, PartialEq, Debug, Marshalable)]
+#[marshalable(tpm2b_simple)]
 pub struct Tpm2bAttest {
     size: u16,
     attestation_data: [u8; size_of::<TpmsAttest>()],
 }
 
 #[repr(C)]
-#[derive(Clone, Copy, PartialEq, Debug)]
+#[derive(Clone, Copy, PartialEq, Debug, Marshalable)]
+#[marshalable(tpm2b_simple)]
 pub struct Tpm2bSymKey {
     size: u16,
     buffer: [u8; TPM2_MAX_SYM_KEY_BYTES as usize],
 }
 
 #[repr(C)]
-#[derive(Clone, Copy, PartialEq, Debug)]
+#[derive(Clone, Copy, PartialEq, Debug, Marshalable)]
+#[marshalable(tpm2b_simple)]
 pub struct Tpm2bLabel {
     size: u16,
     buffer: [u8; TPM2_LABEL_MAX_BUFFER as usize],
@@ -827,7 +837,8 @@ enum TpmuSensitiveCreate {
 }
 
 #[repr(C)]
-#[derive(Clone, Copy, PartialEq, Debug)]
+#[derive(Clone, Copy, PartialEq, Debug, Marshalable)]
+#[marshalable(tpm2b_simple)]
 pub struct Tpm2bSensitiveData {
     size: u16,
     buffer: [u8; TpmuSensitiveCreate::UNION_SIZE],
@@ -843,28 +854,33 @@ pub struct TpmsSensitiveCreate {
 }
 
 #[repr(C)]
-#[derive(Clone, Copy, PartialEq, Debug)]
+#[derive(Clone, Copy, PartialEq, Debug, Marshalable)]
+#[marshalable(tpm2b_simple)]
+#[marshalable(tpm2b_struct)]
 pub struct Tpm2bSensitiveCreate {
     size: u16,
     sensitive: [u8; size_of::<TpmsSensitiveCreate>()],
 }
 
 #[repr(C)]
-#[derive(Clone, Copy, PartialEq, Debug)]
+#[derive(Clone, Copy, PartialEq, Debug, Marshalable)]
+#[marshalable(tpm2b_simple)]
 pub struct Tpm2bPublicKeyRsa {
     size: u16,
     buffer: [u8; TPM2_MAX_RSA_KEY_BYTES as usize],
 }
 
 #[repr(C)]
-#[derive(Clone, Copy, PartialEq, Debug)]
+#[derive(Clone, Copy, PartialEq, Debug, Marshalable)]
+#[marshalable(tpm2b_simple)]
 pub struct Tpm2bPrivateKeyRsa {
     size: u16,
     buffer: [u8; (TPM2_MAX_RSA_KEY_BYTES / 2) as usize],
 }
 
 #[repr(C)]
-#[derive(Clone, Copy, PartialEq, Debug)]
+#[derive(Clone, Copy, PartialEq, Debug, Marshalable)]
+#[marshalable(tpm2b_simple)]
 pub struct Tpm2bEccParameter {
     size: u16,
     buffer: [u8; TPM2_MAX_ECC_KEY_BYTES as usize],
@@ -894,7 +910,8 @@ enum TpmuEncryptedSecret {
 }
 
 #[repr(C)]
-#[derive(Clone, Copy, PartialEq, Debug)]
+#[derive(Clone, Copy, PartialEq, Debug, Marshalable)]
+#[marshalable(tpm2b_simple)]
 pub struct Tpm2bEncryptedSecret {
     size: u16,
     secret: [u8; TpmuEncryptedSecret::UNION_SIZE],
@@ -1149,21 +1166,25 @@ impl Marshalable for TpmtPublic {
 }
 
 #[repr(C)]
-#[derive(Clone, Copy, PartialEq, Debug)]
+#[derive(Clone, Copy, PartialEq, Debug, Marshalable)]
+#[marshalable(tpm2b_simple)]
+#[marshalable(tpm2b_struct)]
 pub struct Tpm2bPublic {
     size: u16,
     public_area: [u8; size_of::<TpmtPublic>()],
 }
 
 #[repr(C)]
-#[derive(Clone, Copy, PartialEq, Debug)]
+#[derive(Clone, Copy, PartialEq, Debug, Marshalable)]
+#[marshalable(tpm2b_simple)]
 pub struct Tpm2bTemplate {
     size: u16,
     buffer: [u8; size_of::<TpmtPublic>()],
 }
 
 #[repr(C)]
-#[derive(Clone, Copy, PartialEq, Debug)]
+#[derive(Clone, Copy, PartialEq, Debug, Marshalable)]
+#[marshalable(tpm2b_simple)]
 pub struct Tpm2bPrivateVendorSpecific {
     size: u16,
     buffer: [u8; TPM2_PRIVATE_VENDOR_SPECIFIC_BYTES as usize],
@@ -1361,7 +1382,8 @@ pub struct _PRIVATE {
 }
 
 #[repr(C)]
-#[derive(Clone, Copy, PartialEq, Debug)]
+#[derive(Clone, Copy, PartialEq, Debug, Marshalable)]
+#[marshalable(tpm2b_simple)]
 pub struct Tpm2bPrivate {
     size: u16,
     buffer: [u8; size_of::<_PRIVATE>()],
@@ -1375,7 +1397,8 @@ pub struct TpmsIdObject {
 }
 
 #[repr(C)]
-#[derive(Clone, Copy, PartialEq, Debug)]
+#[derive(Clone, Copy, PartialEq, Debug, Marshalable)]
+#[marshalable(tpm2b_simple)]
 pub struct Tpm2bIdObject {
     size: u16,
     credential: [u8; size_of::<TpmsIdObject>()],
@@ -1399,7 +1422,8 @@ pub struct Tpm2bNvPublic {
 }
 
 #[repr(C)]
-#[derive(Clone, Copy, PartialEq, Debug)]
+#[derive(Clone, Copy, PartialEq, Debug, Marshalable)]
+#[marshalable(tpm2b_simple)]
 pub struct Tpm2bContextSensitive {
     size: u16,
     buffer: [u8; TPM2_MAX_CONTEXT_SIZE as usize],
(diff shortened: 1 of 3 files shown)


marshalable-derive/src/lib.rs
```diff
@@ -65,25 +94,271 @@ fn derive_tpm_marshal_inner(input: DeriveInput) -> Result<TokenStream> {
...
     Ok(expanded)
 }
 
+fn derive_tpm2b_code(
+    tpm2b_outer_struct_name: &Ident,
+    stru: &syn::DataStruct,
+    tpm2b_simple: bool,
+    tpm2b_struct: bool,
+) -> TokenStream {
+    const TPM2B_SIMPLE_ERR_GENERAL: &str = "A Tpm2b struct must contain just to elements, a u16 size and a buffer array when using #[marshalable(tpm2b_simple)]";
```
oddcoder, line 127: small nit: "two"

oddcoder: This is a top level documentation for how to use `tpm2b_simple`, and not for how to use the code generated by `tpm2b_simple`. So I suggest placing it here https://github.com/tpm-rs/tpm-rs/blob/bd995111e44f04bfbd677080d887672feebf60a7/marshalable-derive/src/lib.rs#L12 with an example if possible.

marshalable-derive/src/lib.rs
```diff
@@ -65,25 +94,271 @@ fn derive_tpm_marshal_inner(input: DeriveInput) -> Result<TokenStream> {
...
+        // Lets validate the second element in the Tpm2b structure and extract name and size expression.
+        let second = &fields.named[1];
+        let field_name = second.ident.as_ref().unwrap();
+        let field_type = &second.ty;
+
+        // Expect array type.
+        let type_array = if let syn::Type::Array(type_array) = field_type {
+            type_array
+        } else {
+            panic!("{}", TPM2B_SIMPLE_ERR_2ND_ELM);
```
oddcoder, line 167: please do not panic, instead propegate errors to top level similar to https://github.com/tpm-rs/tpm-rs/blob/bd995111e44f04bfbd677080d887672feebf60a7/marshalable-derive/src/lib.rs#L62

base/src/lib.rs
```diff
@@ -843,28 +854,33 @@ pub struct TpmsSensitiveCreate {
 }
 
 #[repr(C)]
-#[derive(Clone, Copy, PartialEq, Debug)]
+#[derive(Clone, Copy, PartialEq, Debug, Marshalable)]
+#[marshalable(tpm2b_simple)]
+#[marshalable(tpm2b_struct)]
```
oddcoder, line 859: what is the difference between `tpm2b_struct` and `tpm2b_simple` ?

marshalable-derive/Cargo.toml
```diff
@@ -10,3 +10,5 @@ proc-macro = true
 proc-macro2 = { workspace = true }
 quote = { workspace = true }
 syn = { workspace = true }
+log = "0.4.22"
```
oddcoder, line 13: why are are having those, you do not depend on log at all! also depending on log in proc macro is generally undesirable, rust can already propagate error as compiler error

marshalable-derive/src/lib.rs
```diff
@@ -1,14 +1,28 @@
...
 use quote::{quote, quote_spanned};
+use std::collections::HashMap;
 use syn::{
     parse_macro_input, spanned::Spanned, Attribute, Data, DataEnum, DeriveInput, Error, Expr,
-    Field, Fields, FieldsNamed, Ident, Index, MetaNameValue, Path, Result, Type,
+    ExprCall, ExprPath, Field, Fields, FieldsNamed, Ident, Index, MetaNameValue, Path,
+    PathArguments, Result, Type,
 };
 
+// Enable for logging during compilation. Remember to call "init_logger();" at initialization.
```
oddcoder, line 12: to be removed ?

oddcoder:
I did initial review, with minor comments, aside from the minor modification requested, it will be good idea to have unit tests in https://github.com/tpm-rs/tpm-rs/blob/main/marshalable/src/tests.rs

marshalable-derive/src/lib.rs
TRJensen: Fixed.

marshalable-derive/Cargo.toml
TRJensen: I forgot to remove this!! It was used for dumping syntax tree at compile time, to support during development. Removed again.

marshalable-derive/src/lib.rs
TRJensen: It was used for dumping syntax tree at compile time, to support during development. Removed again.

marshalable-derive/src/lib.rs
TRJensen: Fixed

trent-reed:
Sorry for the delay, I'll take a look at this on Monday.

base/src/lib.rs
TRJensen: The tpm2b_struct and tpm2b_simple is based on the original macro.

I can see that the code added to the 3 structs is the following:
```
            quote! {
                impl Tpm2bStruct for #tpm2b_outer_struct_name {
                    type StructType = #tpm2b_inner_struct_name;

                    fn from_struct(val: &Self::StructType) -> TpmRcResult<Self> {
                        let mut x = Self::default();
                        x.size = val.try_marshal(&mut x.#field_name)? as u16;
                        Ok(x)
                    }

                    fn to_struct(&self) -> TpmRcResult<Self::StructType> {
                        let mut buf = UnmarshalBuf::new(&self.#field_name[0..self.get_size() as usize]);
                        Self::StructType::try_unmarshal(&mut buf)
                    }
                }
            }
```
It is highly dependent on Marshaling. @oddcoder Are you recommending to:
1. Create a new top level folder "tpm2bstruct-derive" or similar. Or,
2. Implement the derive functionality within the existing "marshalable-derive/src"
The derive functionality should confirm that Marshallable is defined the the structs marked with #[derive(Marshalable)] - but I can probably work around that.

trent-reed:
Would you look at that, it's Friday! Busier week than I expected - sorry about that... I'll take a look as soon as I am able to next week.

base/src/lib.rs
oddcoder: I mean option 2, and maybe we need to rename it from marshalable-derive to something else, but it should not matter for now.

> It is highly dependent on Marshaling

Note that in rust the general pattern when it comes to derive macros, is that one macro derives one trait, You can look at `Copy` and `Clone` for instance, both trait are so closely related, but deriving copy does not automatically derive clone.

>  The derive functionality should confirm that Marshallable is defined the the structs marked with #[derive(Marshalable)] - but I can probably work around that.

Not It does not, the compiler will complain that `x.foo_bar(arg1, arg2)` exists, but it is part of if trait `FooBar` which is not implmented for the right type. you don't really have to do anything.


I will go through review again once it is separated, or if you have better idea.

TRJensen:
> Would you look at that, it's Friday! Busier week than I expected - sorry about that... I'll take a look as soon as I am able to next week.

No problem at all - I have been busy as well :-)

trent-reed:
Okay... Sorry for the huge delay. This is also something I noticed a while back, and just haven't had time to share my thoughts on it. TL;DR, I don't even think we need a macro, but if we do need a macro I'd want it implemented slightly differently from how this PR is trying to define it.

See my comment here in the original issues for detail: https://github.com/tpm-rs/tpm-rs/issues/114#issuecomment-2669402352

oddcoder:
Sorry for the late response, my on-boarding right now at the new job is like a roller coaster (it was explained to me as drinking from fire hydrant), and I am scrapping for free time.

> if we do need a macro I'd want it implemented slightly differently from how this PR is trying to define it.

So The idea is we should not be comparing "different trait + macro to the current trait + derive". Imo the way to go is comparing the proposed trait + macro to not having anything until we have them.

I believe turning the `impl_try_marshalable_tpm2b_simple` into a proper derive is defintely improvement, and it is almost done, so given the little resources we have, "good enough" should  be sufficient (except for implementing cryptography) and we should see what to prioritize next.


@TRJensen if you addressed my last request regarding splitting the derive into two separate derive let me know (also @ me) I can allocate time to look at it over the weekend.

marshalable-derive/src/lib.rs
```diff
@@ -65,25 +80,278 @@ fn derive_tpm_marshal_inner(input: DeriveInput) -> Result<TokenStream> {
...
+            }
+        };
+
+        // Lets validate the second element in the Tpm2b structure and extract name and size expression.
+        let second = &fields.named[1];
+        let field_name = second.ident.as_ref().unwrap();
+        let field_type = &second.ty;
+
+        // Expect array type.
+        let type_array = if let syn::Type::Array(type_array) = field_type {
```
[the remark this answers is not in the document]
TRJensen: Nice & elegant - fixed 👍

marshalable-derive/src/lib.rs
```diff
@@ -65,25 +80,278 @@ fn derive_tpm_marshal_inner(input: DeriveInput) -> Result<TokenStream> {
...
+            type_array
+        } else {
+            return Err(Error::new(field_type.span(), TPM2B_SIMPLE_ERR_2ND_ELM));
+        };
+
+        // Extract the type used.
+        let type_path = if let syn::Type::Path(type_path) = &*type_array.elem {
```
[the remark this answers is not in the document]
TRJensen: Nice - Type was already added - fixed 4 locations 👍

marshalable-derive/src/lib.rs
```diff
@@ -65,25 +80,278 @@ fn derive_tpm_marshal_inner(input: DeriveInput) -> Result<TokenStream> {
...
+            return Err(Error::new(type_array.elem.span(), TPM2B_SIMPLE_ERR_2ND_ELM));
+        };
+
+        // Lets extract the size expression.
+        let size_expression = &type_array.len;
+
+        // Lets extract the <some struct> name from tpm2b buffer definition <buffer name>: [u8, sizeof( <some struct>) ]
+        let tpm2b_inner_struct_name = if tpm2b_struct {
+            // Extract the func, args.
+            let func = if let Expr::Call(ExprCall { func, .. }) = size_expression {
```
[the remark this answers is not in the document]
TRJensen: Check 👍

marshalable-derive/src/lib.rs
```diff
@@ -65,25 +80,278 @@ fn derive_tpm_marshal_inner(input: DeriveInput) -> Result<TokenStream> {
...
+                func
+            } else {
+                return Err(Error::new(size_expression.span(), TPM2B_SIMPLE_ERR_2ND_ELM));
+            };
+
+            // Check if the function being called is "sizeof"
+            let tpm2b_inner_struct_name = if let Expr::Path(ExprPath { path, .. }) = &**func {
```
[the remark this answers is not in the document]
TRJensen: Nice, again 👍 and restructured 👍

marshalable-derive/src/lib.rs
```diff
@@ -65,25 +80,278 @@ fn derive_tpm_marshal_inner(input: DeriveInput) -> Result<TokenStream> {
...
+                // The path contains multiple elements, the first must be sizeof.
+                let segment = path.segments.first().expect(TPM2B_STRUCT_ERR_2ND_ELM);
+                if segment.ident != "size_of" {
+                    return Err(Error::new(path.segments.span(), TPM2B_SIMPLE_ERR_2ND_ELM));
+                };
+                let arguments =
+                    if let PathArguments::AngleBracketed(sizeof_args) = &segment.arguments {
```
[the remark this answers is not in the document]
TRJensen: Check 👍

marshalable-derive/src/lib.rs
```diff
@@ -65,25 +80,278 @@ fn derive_tpm_marshal_inner(input: DeriveInput) -> Result<TokenStream> {
...
+                        type_path
+                    } else {
+                        return Err(Error::new(struct_type.span(), TPM2B_SIMPLE_ERR_2ND_ELM));
+                    };
+
+                let struct_ident = struct_path
+                    .path
+                    .segments
+                    .first()
+                    .expect(TPM2B_STRUCT_ERR_2ND_ELM);
```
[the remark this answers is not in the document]
TRJensen: Ah - I got it - expect() will panic, so I should do the "assign with ? mark" instead to return a clean error.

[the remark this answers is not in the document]
TRJensen: Fixed - I wanted to control the error message, so I used .ok_or() instead of expect (panic). Let me know if this fits your initial thought as well.

marshalable-derive/src/lib.rs
```diff
@@ -9,6 +9,126 @@ use syn::{
...
+    const TPM2B_SIMPLE_ERR_2ND_ELM: &str = "Second element in Tpm2b struct must be u8 buffer: <buffer name>: [u8, <buffer size>] when using #[derive(Tpm2bStruct)].";
+    const TPM2B_STRUCT_ERR_2ND_ELM: &str = "Second element in Tpm2b struct must define a u8 buffer using sizeof expression: <buffer name>: [u8, sizeof( <some struct>) ] when using #[derive(Tpm2bStruct)].";
+
+    // Lets extracts the 2nd element u8 field name (&Ident) and its size expression (&Expr).
+    let (field_name, size_expression) = derive_tpm2b_common(tpm2b_outer_struct_name, stru)?;
+
+    // Lets extract the <some struct> name from tpm2b buffer definition <buffer name>: [u8, sizeof( <some struct>) ]
+    let tpm2b_inner_struct_name = {
+        // Extract the func, args.
+        let func = if let Expr::Call(ExprCall { func, .. }) = size_expression {
```
trent-reed, line 64: (Here and below) Just in general, if you find yourself doing:

```rs
let foo = if let SomeMatch(...) = bar { ... } else { return Err(...); };
```

you should almost always instead:

```rs
let SomeMatch(...) = bar else { return Err(...); };
```

marshalable-derive/src/lib.rs
```diff
@@ -9,6 +9,126 @@ use syn::{
...
+            func
+        } else {
+            return Err(Error::new(size_expression.span(), TPM2B_SIMPLE_ERR_2ND_ELM));
+        };
+
+        // Check if the function being called is "sizeof"
+        let tpm2b_inner_struct_name = if let Expr::Path(ExprPath { path, .. }) = &**func {
+            // The path contains multiple elements, the first must be sizeof.
+            let segment = path.segments.first().expect(TPM2B_STRUCT_ERR_2ND_ELM);
```
trent-reed, line 73: (Here and below) `expect` will panic, please make this a result and handle the error appropriately.

marshalable-derive/src/lib.rs
```diff
@@ -328,6 +378,7 @@ fn derive_tpm2b_code(
     }
 }
 
+
```
trent-reed, line 381: Strange latent newline?

trent-reed (commented):
This new commit seems incomplete, was I supposed to review this or should I wait?

marshalable-derive/src/lib.rs
```diff
@@ -65,25 +80,278 @@ fn derive_tpm_marshal_inner(input: DeriveInput) -> Result<TokenStream> {
...
+    if !tpm2b_simple {
+        return Err(Error::new(
+            tpm2b_outer_struct_name.span(),
+            "We can't have #[marshalable(tpm2b_struct)] without #[marshalable(tpm2b_simple)]",
+        ));
+    }
+
+    if let Fields::Named(ref fields) = stru.fields {
+        // First make sure we have precisely 2 elements:
+        if fields.named.len() != 2 {
```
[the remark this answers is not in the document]
TRJensen: The Punctuated<Field, Comma>, did not provide as_slice(), but I found that I can call fields.named.iter().collect(), and use as_slice() on this one instead.

marshalable-derive/src/lib.rs
```diff
@@ -65,25 +80,278 @@ fn derive_tpm_marshal_inner(input: DeriveInput) -> Result<TokenStream> {
...
+            return Err(Error::new(fields.named.span(), TPM2B_SIMPLE_ERR_GENERAL));
+        }
+
+        // Lets validate the first element is named size and has type u16.
+        {
+            let first = &fields.named[0];
+            let field_name = first.ident.as_ref().unwrap();
```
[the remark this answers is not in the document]
TRJensen: Fixed (in next upload).

marshalable-derive/src/lib.rs
```diff
@@ -65,25 +80,278 @@ fn derive_tpm_marshal_inner(input: DeriveInput) -> Result<TokenStream> {
...
+        // Confirm it is an array of u8 elements.
+        if !type_path.path.is_ident("u8") {
+            return Err(Error::new(type_array.elem.span(), TPM2B_SIMPLE_ERR_2ND_ELM));
+        };
+
+        // Lets extract the size expression.
+        let size_expression = &type_array.len;
+
+        // Lets extract the <some struct> name from tpm2b buffer definition <buffer name>: [u8, sizeof( <some struct>) ]
+        let tpm2b_inner_struct_name = if tpm2b_struct {
```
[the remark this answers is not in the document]
TRJensen: It has been restructured from previous comments (next upload)

marshalable-derive/src/lib.rs
```diff
@@ -9,6 +9,126 @@ use syn::{
...
+pub fn derive_tpm_2b_struct(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
+    let input = parse_macro_input!(input as DeriveInput);
+    match derive_tpm_2b_struct_inner(input) {
+        Ok(t) => t.into(),
+        Err(e) => e.to_compile_error().into(),
+    }
+}
+
+fn derive_tpm_2b_struct_inner(input: DeriveInput) -> Result<TokenStream> {
+    let name = input.ident;
```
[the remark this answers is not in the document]
TRJensen: Fixed (in next upload)

marshalable-derive/src/lib.rs
TRJensen: Fixed (in next upload)

marshalable-derive/src/lib.rs
TRJensen: Fixed (in next upload)

marshalable-derive/src/lib.rs
TRJensen: Added rust comment for the derive, however kept my const string for returning explaining error string.

TRJensen:
@oddcoder @trent-reed : I have now uploaded a new revised patch (after some iterations) - and requested a re-review.
I hope I got all findings.

@trent-reed Sorry about the comments with no immediate patch upload - I was commenting as I fixed the code. Somehow I am missing the gitlab feature, where it is possible to draft multiple comments, and first publish them after the patch is pushed 🙂 

Thanks for the great learning experience and the many great tips and tricks👍🙂 Let me know if I missed something or any other great ideas 👍

marshalable-derive/src/lib.rs
```diff
@@ -9,6 +9,108 @@ use syn::{
...
+
+    let Some(struct_type) = arguments.args.first() else {
+        return Err(Error::new(segment.span(), TPM2B_STRUCT_ERR_2ND_ELM));
+    };
+
+    let syn::GenericArgument::Type(Type::Path(struct_path)) = struct_type else {
+        return Err(Error::new(struct_type.span(), TPM2B_STRUCT_ERR_2ND_ELM));
+    };
+
+    let tpm2b_inner_struct_name = struct_path.path.segments.first();
```
trent-reed, line 89: Probably you just want to use `struct_path.path`, which I think you can do:

```rs
let tpm2b_inner_struct_name = &struct_path.path;
```

This means if the user uses a type like `foo::Bar`, then codegen will do something like:

```rs
impl Tpm2bStruct for <name> {
  type StructType = foo::Bar;

  // ...
}
```

Which is probably what you want anyways, right?

trent-reed (approved):
Only nits remain, thanks!

marshalable-derive/src/lib.rs
```diff
@@ -9,6 +9,108 @@ use syn::{
...
+            segment.arguments.span(),
+            TPM2B_STRUCT_ERR_2ND_ELM,
+        ));
+    };
+
+    let Some(struct_type) = arguments.args.first() else {
+        return Err(Error::new(segment.span(), TPM2B_STRUCT_ERR_2ND_ELM));
+    };
+
+    let syn::GenericArgument::Type(Type::Path(struct_path)) = struct_type else {
```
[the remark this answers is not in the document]
TRJensen: Done 👍

marshalable-derive/src/lib.rs
```diff
@@ -9,6 +9,108 @@ use syn::{
...
+
+                fn to_struct(&self) -> TpmRcResult<Self::StructType> {
+                    let mut buf = UnmarshalBuf::new(&self.#field_name[0..self.get_size() as usize]);
+                    Self::StructType::try_unmarshal(&mut buf)
+                }
+            }
+        };
+        Ok(quote! { #tpm2b_struct_code })
+    } else {
+        Ok(quote! {})
```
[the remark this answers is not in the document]
TRJensen: It disappeared - Done 👍

marshalable-derive/src/lib.rs
```diff
@@ -9,6 +9,108 @@ use syn::{
...
+    };
+
+    let syn::GenericArgument::Type(Type::Path(struct_path)) = struct_type else {
+        return Err(Error::new(struct_type.span(), TPM2B_STRUCT_ERR_2ND_ELM));
+    };
+
+    let tpm2b_inner_struct_name = struct_path.path.segments.first();
+
+    if tpm2b_inner_struct_name.is_some() {
+        let tpm2b_struct_code = quote! {
```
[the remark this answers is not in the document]
TRJensen: Done 👍

marshalable-derive/src/lib.rs
```diff
@@ -9,6 +9,108 @@ use syn::{
...
+            input.span(),
+            "The #[derive(Tpm2bStruct)] can only be applied to a struct.",
+        )),
+    }
+}
+
+fn derive_tpm2b_struct(
+    tpm2b_outer_struct_name: &Ident,
+    data_struct: &syn::DataStruct,
+) -> Result<TokenStream> {
```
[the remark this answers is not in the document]
TRJensen: Noted 👍

marshalable-derive/src/lib.rs
TRJensen: Correct - Done 👍

TRJensen:
@trent-reed Thank you for the final nits - I have fixed and merged👍 🙂

PR #152, opened 2025-03-05
commands: Add CreatePrimary marshalling

deeglaze:
I'm very new to Rust, so I well expect this to not be what y'all would prefer. I do hope this is around about what y'all plan to do with commands and responses.

If this looks right stylistically, I'll go ahead and add some unit tests.

### DIFF
--- a/base/src/commands/hierarchy.rs
+++ b/base/src/commands/hierarchy.rs
@@ -1,7 +1,38 @@
 //! [TPM2.0 1.83] 24 Hierarchy Commands
 
+use crate::{commands::*, *};
+
+// Defined by TCG EK Credential Profile 2.0.
+// This is the PolicyOR of PolicyA and PolicyC.
+// PolicyA is empty digest | PolicySecret | endorsement hierarchy.
+// PolicyC is empty digest | PolicyAuthorizeNV | Name of index I-1
+// Index I-1 is 0x01C07F01 with name alg SHA256, auth policy PolicyASha256,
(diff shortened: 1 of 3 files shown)


deeglaze:

base/src/commands/hierarchy.rs
```diff
@@ -23,3 +42,41 @@ pub struct ClearControlCmd {}
...
+	    sensitive: TpmsSensitiveCreate::default(),
+	    public: TpmtPublic{
+		name_alg: TpmiAlgHash::SHA256,
+		object_attributes: TpmaObject(0x300b2),
+		auth_policy: Tpm2bDigest{
+		    size: 32,
+		    buffer: {
+			let mut base = [0; 64];
+			base.copy_from_slice(&[
+			    0x83, 0x71, 0x97, 0x67, 0x44, 0x84, 0xb3, 0xf8, 0x1a, 0x90, 0xcc, 0x8d, 0x46, 0xa5,
```
oddcoder, line 59: what are those?

oddcoder:
> @esmusick Since you added the TpmCommand trait, I think maybe you have the best sensibility for how it should be used?

Sadly she is no longer active in the project, will it be possible to add unit tests for that. we have integraton with ms tpm simulator.

base/src/commands/hierarchy.rs
deeglaze: Clarified with a constant.

PR #157, opened 2025-03-06
Add command marshaling support for NVReadPublic*

deeglaze:


deeglaze:
> @deeglaze I started CI, can you please add comments (maybe in commit message, or in the commit where those constants come from. it should make review process easier.
>

The constants are from the sections that the enums' comments point to.

```
// The TPM_CC command codes
// See definition in Part 2: Structures, section 6.5.2.
```
For TPM_CC_NV_ReadPublic2 for example


```
// TpmHt represents a TPM_HT.                                                                                                                                                                                                                                                                                                                        
// See definition in Part 2: Structures, section 7.2. 
```
For ExternalNV and PermanentNV.

deeglaze:
This still needs work with simulation testing. I'm still looking for a reliable way to ensure some NVIndex pre-populated in the simulated environment, such as the low range RSA 2048 EK certificate 0x01c00002 or EK template 0x01c00004

PR #161, opened 2025-04-19
Add support for generating rustdocs for the main branch.

nnooney:
The `docs.yml` workflow publishes the main branch to the GitHub pages site for the repo. It depends on 4 GitHub-published actions:

- https://github.com/actions/checkout
- https://github.com/actions/configure-pages
- https://github.com/actions/upload-pages-artifact
- https://github.com/actions/deploy-pages

The action can be triggered manually in addition to every PR that merges into main.


nnooney:
Before merging this PR, I need a project admin -- @chrisfenner -- to update the Pages settings of this repository to ensure that the GitHub Pages Source is set to `GitHub Actions`; this ensures that the workflow can publish to the local pages.

https://docs.github.com/en/pages/getting-started-with-github-pages/configuring-a-publishing-source-for-your-github-pages-site#publishing-with-a-custom-github-actions-workflow

trent-reed:
> Before merging this PR, I need a project admin -- @chrisfenner -- to update the Pages settings of this repository to ensure that the GitHub Pages Source is set to `GitHub Actions`; this ensures that the workflow can publish to the local pages.
> 
> https://docs.github.com/en/pages/getting-started-with-github-pages/configuring-a-publishing-source-for-your-github-pages-site#publishing-with-a-custom-github-actions-workflow

Awaiting @chrisfenner to do this step before reviewing more thoroughly (added him as a reviewer).

.github/workflows/docs.yml
```diff
@@ -0,0 +1,65 @@
...
+      - name: Build Docs
+        id: build
+        run: cargo doc --no-deps
+
+        # This step is needed because rustdoc doesn't generate a main landing
+        # page. The approach here is to redirect the main page to one of the
+        # crate index pages.
+      - name: Create Index Redirect
+        # Change the 'url' field to choose the default crate to show.
+        run: echo '<meta http-equiv="refresh" content="0; url=tpm2_rs_base/index.html">' > target/doc/index.html
```
trent-reed, line 41: Can users still access the other crates as well? I wonder if we should do something like have a landing page that leads to the other crates, or if this is sufficient?

Maybe I'm thinking too hard about it, WDYT?

nnooney: Yes, near the bottom of the left-hand navigation menu in the generated docs is a list of all of the crates in the project, so it's easy to switch between them. I confirmed this locally by running `cargo doc --no-deps`, which is the same command that the CI runs.

This setting only controls which crate we show by default; I chose `tpm2_rs_base` since that seems like the appropriate lowest-level dependency crate.

chrisfenner:
> Before merging this PR, I need a project admin -- @chrisfenner -- to update the Pages settings of this repository to ensure that the GitHub Pages Source is set to `GitHub Actions`; this ensures that the workflow can publish to the local pages.
> 
> https://docs.github.com/en/pages/getting-started-with-github-pages/configuring-a-publishing-source-for-your-github-pages-site#publishing-with-a-custom-github-actions-workflow

I believe I have set the setting:

image

.github/workflows/docs.yml
trent-reed: Thanks for confirming!

trent-reed (approved):
Approved as long as the setting Chris has changed is the one you wanted changed. Thanks!

PR #166, opened 2025-04-27
Follow module naming conventions outlined in Rust docs

nnooney:
https://doc.rust-lang.org/reference/items/modules.html#r-items.mod.outlined

The docs prefer to use the `module.rs` convention rather than `module/mod.rs` to avoid having too many files with the name `mod.rs` in the filesystem. This change moves all `mod.rs` files up one directory and uses their module name.

There's one exception to this: the simulator_tests_registration module under client/tests. This is because the tests folder treats each file as a separate crate.

https://doc.rust-lang.org/book/ch11-03-test-organization.html#integration-tests


trent-reed:
In my experience, the best strategy is:

1. If there's a module with no submodules, it may be named `<module>.rs`.
2. If there's a submodule, it's best to organize it like `<module>/mod.rs` if you can.
3. Only use `<module>.rs` with additional submodules if you must.

That said, I don't actually care that much.

There was a discussion about this, but looking at our codestyle document, I don't actually see this reflected in the document: https://github.com/tpm-rs/docs/blob/main/coding-style.md

We originally voted on repo organization structure here: https://github.com/orgs/tpm-rs/discussions/19

At that time it used exactly the method of organization you mention in this PR.

However, I think eventually it was discussed further and decided that it was annoying having the big separation between modules and submodules... I think this was discussed in this (unaccepted) PR: https://github.com/tpm-rs/tpm-rs/pull/76

So maybe our coding-styles document needs updated if we still agree with that direction.

trent-reed:
I think the follow-up here is:

1. Update `coding-style.md` with what I outlined above.
2. Update the rest of the code to follow the defined coding style advice.

PR #168, opened 2025-05-28
Implement Tpm trait for TpmSim

### LINKED ISSUE #139: `TpmSim` should implement `Tpm` trait, so that we don't need to hold on to a lifetime of the simulator separately from the `Tpm` client.
Noticed in https://github.com/tpm-rs/tpm-rs/pull/137, 

currently we have to do something like this:

```rs
let (_sim_lifeline, mut tpm) = get_started_tpm();
```

This is a bit odd, just want to open a cleanup bug to where the simulator itself implements `Tpm`. That way we don't need this other separate lifetime.

kstruczy:
Currently to use the simulator client holds on to two different lifetimes, one of the simulator process, and one of the communication (tpm) channel:
let (_sim_lifeline, mut tpm) = get_started_tpm();

To fix this:
- move communication channels to the simulator struct (TpmSim)
- implement Tpm trait on TpmSim
- fix tests using the simulator

This patch adds scopeguard crate to workspace (no-std configuration).

fixes #139

### DIFF
--- a/client/tests/simulator_tests.rs
+++ b/client/tests/simulator_tests.rs
@@ -6,7 +6,7 @@
 /// command line.
 ///
 /// To run locally in the docker container:
-///  `cd client && docker compose run simulator_tests`
+///  `cd client && docker compose run --rm simulator_tests`
 ///
 /// These tests must be run with `--test-threads=1`, because they use a single TCP port.
 use std::io::{Error, ErrorKind, IoSlice, Read, Result, Write};
@@ -36,7 +36,7 @@ fn get_simulator_path() -> String {
 
 #[test]
 fn test_startup_tpm() {
-    let (_sim_lifeline, mut tpm) = run_tpm_simulator(&get_simulator_path()).unwrap();
+    let mut tpm = run_tpm_simulator(&get_simulator_path()).unwrap();
     let startup = StartupCmd {
         startup_type: TpmSu::Clear,
     };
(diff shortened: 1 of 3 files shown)


ghost (changes requested):
Sorry for the huge delay.

client/tests/simulator_tests.rs
```diff
@@ -36,65 +37,133 @@ fn get_simulator_path() -> String {
...
+
+    // Starts up the TPM simulator with a platform server listening at the give IP/port.
+    fn start_tcp_tpm(&mut self) -> Result<()> {
+        TpmCommand::SignalNvOff.issue_to_platform(&mut self.plat_channel.0)?;
+        TpmCommand::SignalPowerOff.issue_to_platform(&mut self.plat_channel.0)?;
+        TpmCommand::SignalPowerOn.issue_to_platform(&mut self.plat_channel.0)?;
+        TpmCommand::SignalNvOn.issue_to_platform(&mut self.plat_channel.0)
+    }
+}
+impl Tpm for TpmSim {
```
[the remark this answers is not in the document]
kstruczy: I agree, good to fix it in the future but I don't think it's in the scope of this PR.

client/tests/simulator_tests.rs
```diff
@@ -127,55 +196,39 @@ struct TcpTpmHeader {
     locality: u8,
     cmd_len: U32,
 }
-// Provides TCP transport for talking to a TPM simulator.
-pub struct TcpTpm {
-    tpm_conn: TcpStream,
+
+struct TcpChannel(TcpStream);
```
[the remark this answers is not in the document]
kstruczy: Yeah, I wanted to have a different abstraction for "platform" and "tpm" channel, but I agree it doesn't look great and complicates things. I'll fix it the way you suggested, thanks.

Not in the scope of this PR, but I'm wondering if it's not better to separate "platform" and "tpm" commands, considering that they have separate channels and are handled differently.

Cargo.toml
```diff
@@ -19,6 +19,7 @@ open-enum = "0.4.1"
 proc-macro2 = "1"
 quote = "1"
 safe-discriminant = "0.2.0"
+scopeguard = { version = "1.2.0", default-features = false }
```
oddcoder, line 22: I do not understand why do we need this new dependency. Can't we do without it?

oddcoder:
I am to great extent in favor of using drop trait instead of adding a new
dependeny just for wrapper around drop trait.

It will be at most 10 lines of code.

client/tests/simulator_tests.rs
```diff
@@ -36,65 +37,139 @@ fn get_simulator_path() -> String {
...
-            Command::new(format!(".{simulator_bin}"))
-                .current_dir("/")
-                .spawn()?,
-        ))
+    fn new(simulator_bin: &str, ip: &str, plat_port: u16, tpm_port: u16) -> Result<TpmSim> {
+        let child = Command::new(format!(".{simulator_bin}"))
+            .current_dir("/")
+            .spawn()?;
+
+        let child_guard = scopeguard::guard(child, |mut child| {
```
oddcoder, line 89:
```rust
impl Drop for TpmSim {
   fn drop (self) {
     //do same stuff to self.child
   }
}
```

client/tests/simulator_tests.rs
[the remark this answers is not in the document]
trent-reed: Correct me if I'm wrong, but I thought the platform address still handled something that looks like a TPM command, just that there's other commands available. I could be misremembering though. Anyways, I'm not sure, we can probably handle that as we go.

Were you able to make this simplification, or were there issues with it?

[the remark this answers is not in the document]
kstruczy: No issues, please have a look at the new version that will also remove the scopeguard dependency.

client/tests/simulator_tests.rs
```diff
@@ -36,63 +36,150 @@ fn get_simulator_path() -> String {
...
+
 impl TpmSim {
-    fn new(simulator_bin: &str) -> Result<TpmSim> {
-        Ok(TpmSim(
-            Command::new(format!(".{simulator_bin}"))
-                .current_dir("/")
-                .spawn()?,
-        ))
+    fn new(simulator_bin: &str, ip: &str, plat_port: u16, tpm_port: u16) -> Result<TpmSim> {
+        let child = Command::new(format!(".{simulator_bin}"))
```
trent-reed, line 84: In a later CL; I know you didn't add this, but I find this format with the dot weird. Normally we should just trust that the input points to the correct path. Maybe a small fix is needed here. I'll file a small tracking item: https://github.com/tpm-rs/tpm-rs/issues/173

client/tests/simulator_tests.rs
```diff
@@ -36,63 +36,150 @@ fn get_simulator_path() -> String {
...
+            .current_dir("/")
+            .spawn()?;
+
+        let child_guard = ChildGuard(Some(child));
```
trent-reed, line 88: Nit: I'd rather you just provide a constructor for this, personally.

```rs
ChildGuard::new(child)
```

You'll almost never want to build one of these without having a child to populate into it.

client/tests/simulator_tests.rs
```diff
@@ -36,63 +36,150 @@ fn get_simulator_path() -> String {
...
+
+        let plat_channel = TcpTpmChannel::new(ip, plat_port)?;
+        let tpm_channel = TcpTpmChannel::new(ip, tpm_port)?;
+
+        let child = child_guard.into_inner();
```
trent-reed, line 93: Why not just keep the drop guard so that it kills the process when it goes out of scope? I would not suggest this `into_inner` function, and instead just hold on to the `ChildGuard` in `TpmSim`.

You anyways do actually want the process to be killed when the `TpmSim` goes out of scope, which is what `ChildGuard` provides.

client/tests/simulator_tests.rs
```diff
@@ -36,63 +36,150 @@ fn get_simulator_path() -> String {
...
-fn start_tcp_tpm(ip: &str, plat_port: u16) -> Result<()> {
-    let mut connection = TcpStream::connect(format!("{ip}:{plat_port}"))?;
-    TpmCommand::SignalNvOff.issue_to_platform(&mut connection)?;
-    TpmCommand::SignalPowerOff.issue_to_platform(&mut connection)?;
-    TpmCommand::SignalPowerOn.issue_to_platform(&mut connection)?;
-    TpmCommand::SignalNvOn.issue_to_platform(&mut connection)
+struct ChildGuard(Option<Child>);
+
+impl ChildGuard {
+    fn into_inner(mut self) -> Child {
```
trent-reed, line 162: As stated above, I'd remove this.

client/tests/simulator_tests.rs
```diff
@@ -36,63 +36,150 @@ fn get_simulator_path() -> String {
...
+        self.0.take().expect("Child already dropped")
+    }
+}
+impl Drop for ChildGuard {
+    fn drop(&mut self) {
+        if let Some(mut child) = self.0.take() {
```
trent-reed, line 168: Nit: Could be simplified:

```
let Some(mut child) = self.0.take() else { return };
child.kill().inspect_err(|err|
    eprintln!("Failed to kill simulator child process after setup failure: {err}"
));
child.wait().inspect_err(|err|
    eprintln!("Failed to wait for simulator child process after setup failure: {err}"
));
```

client/tests/simulator_tests.rs
```diff
@@ -36,63 +36,150 @@ fn get_simulator_path() -> String {
...
+            if let Err(kill_err) = child.kill() {
+                eprintln!(
+                    "Failed to kill simulator child process after setup failure: {}",
+                    kill_err
+                );
+            }
+            if let Err(wait_err) = child.wait() {
```
trent-reed, line 175: Do we need this if we just killed it above? I think not, right?

Issue #97, opened 2024-08-28
Agree on the `tpm2-rs-base` public module topology


trent-reed:
## Summary

There's a number of ways we could organize this crate, we should agree on what we want the final topology to look like. (Note: This has to do with the publicly visible modules, not the private modules with public re-exports for code organization.)

## Details

As a part of [Discussion #18](https://github.com/orgs/tpm-rs/discussions/18), I've been slowly working on breaking up these monolithic modules. However, there's one last thing we haven't really discussed - and that's the ideal module topology from the perspective of the client.

Keep in mind, the developer module topology (what we see and maintain) does not have to match what the client sees. We can have a completely flat repo from the perspective of the client, but separate into submodules internally and simply re-export.

With that said, I've seen that it's usually nice to expose some kind of a structure to the client, that way there's some sense to the crate layout, and they aren't bombarded with types when they look at the documentation.

We agree that we should break these modules up, but I don't think we agree on what the ideal final state of the modules are from the perspective of the client.

### A Note on Enums in the `constants` Module

I highly suspect that these are only named "constants" in the spec because the folks who penned the spec have the perspective of C programmers.

Enums are constants, according to C. So it makes some kind os sense to say "these types are constants". But from the perspective of rust, they're just `enum` types (well, because of `open_enum`, they're structures with constant `impl` values that look like `enums`, but let's not get too in the weeds over that).

My feeling is these should be treated like other types. So they should be named similarly (`PascalCase`), and if we decide to organize types into a module named `types`, they actually belong there - not in the `constants` module.

If we keep a `constants` module, I think we should literally only have constants in there.

### A Note on Prefix Name Stripping

This has come up a few times ([Issue #65](https://github.com/tpm-rs/tpm-rs/issues/65), [Issue #15](https://github.com/tpm-rs/tpm-rs/issues/15)).

My opinion is that we don't want to separate the prefixes from the type names, though I'm open to discussing this more if folks disagree. I'll copy my reasons for this from [Issue #15](https://github.com/tpm-rs/tpm-rs/issues/15):

[...]

PR #192, opened 2025-07-25
WIP: Add dedicated MarshalingError for marshal/unmarshal functionality.

TRJensen:
A dedicated marshaling error, gives are more precise indication of if the marshaling failed on TPM or on the TPM client. For now MarshallingError is used for both unmarshal/marshal.


errors/src/lib.rs
```diff
@@ -37,3 +39,57 @@ impl TryFrom<u32> for TssError {
         }
     }
 }
+
+/// Represents success or [`MarshalingError`] failure, which is used for unmarshal/marshal functionality.
+pub type MarshalingResult<T> = Result<T, MarshalingError>;
```
trent-reed, line 44: Ahh, naming...

When I originally prefixed the type names with `Marshaling*` in the issue it was because there was another type `Unmarshaling*` that it needed to be disambiguated from. If we are squashing this down to one type, it's possible we don't need the naming disambiguation and we can disambiguate based on submodule location.

## Submodule Disambiguation

If we had a submodule `marshaling` and under that a type named `Result` which was the alias to `Result<T, marshaling::Error>`, that would be a disambiguation by submodule.

As mentioned in the other comments, today this would be in `tpm2-marshaling`, but in the future I'd envision it as a submodule of `tpm2-base`, since really we want to get rid of some of these crates.

## Naming Woes With TPM Types

I'm actually not opposed to naming things plainly like this in Rust.

I just thought we should maybe avoid it for the TPM types since those are named in the spec. From the latest meeting though, @josephlr seems to disagree with me on this and is trying to open that back up for discussion - I don't actually mind that much, it's not a hill I'm willing to die on, there's pros and cons each way, so we just need to pick one.

So if we want we can name the TPM types plainly like `tpm2b::Whatever` as well that's fine - one benefit, as described in #97, is that we can put the documentation for specific type patterns someplace sensible. I don't know if these two conversations intersect, but they seem like they might, so paging @josephlr to comment.

errors/src/lib.rs
```diff
@@ -37,3 +39,57 @@ impl TryFrom<u32> for TssError {
...
+            MarshalingError::UnknownSelector => TpmRcError::Selector.into(),
+            MarshalingError::MarshalingDeriveError => TpmRcError::Failure.into(),
+        }
+    }
+}
+
+// The MarshalingError defines Unmarshaling/Marshaling errors codes,
+// providing more explicit error codes for try_marshal* and try_unmarchal*.
+#[derive(PartialEq, Eq, Copy, Clone, Debug)]
+pub enum MarshalingError {
```
trent-reed, line 90: Strange to define `MarshalingError` after your impl blocks, I would rather this be before the impl blocks.

errors/src/lib.rs
```diff
@@ -37,3 +39,57 @@ impl TryFrom<u32> for TssError {
         }
     }
 }
+
```
trent-reed, line 42: These changes belong in `marshaling` crate (which as we discussed might in the future be removed, but for now that's where these changes belong). These are errors related to marshaling, and as mentioned the only reason this separate errors crate exists is because the marshaling error type does not. So what will probably happen here is:

1. You implement this/these error types in `tpm2-marshaling` (`From` conversions also in `marshaling`).
2. We move `TpmError` to `tpm2-base`, removing the `tpm2-error` crate (`From` conversions now belong in `base`).
3. Then we remove the `tpm2-marshaling` crate and rename `tpm2-marshaling-derive` to just `tpm2-derive`, moving the contents of `tpm2-marshaling` to `tpm2-base`.
4. Eventually we also remove the `tpm2-unionify` crate, but that's a project for another day.

Paging @josephlr to comment.

errors/src/lib.rs
trent-reed, line 90: I originally proposed this as a separate type, @oddcoder wanted a single type, and @josephlr still seemed to want separate types? I don't care which any more, but please figure out which direction makes the most sense and everyone agrees on the most here. Maybe @josephlr should state his thoughts on this since it's the latest comment on the matter.

Do note, if they are separate types the naming becomes a bit more complicated; as described in the other comments if it's one type we can name the alias `marshaling::Result` and `marshaling::Error`, if it's separate types we either have to make separate submodules (`marshaling::Error` vs `unmarshaling::Error`, etc), OR we will actually need to prefix the type names (`MarshalingError` vs `UnmarshalingError`, etc).

errors/src/lib.rs
```diff
@@ -37,3 +39,57 @@ impl TryFrom<u32> for TssError {
...
+        match orig {
+            MarshalingError::ArrayLengthExceeded => TpmRcError::Size.into(),
+            MarshalingError::UnexpectedEndOfBuffer => TpmRcError::Memory.into(),
+            MarshalingError::UnknownSelector => TpmRcError::Selector.into(),
+            MarshalingError::MarshalingDeriveError => TpmRcError::Failure.into(),
+        }
+    }
+}
+
+// The MarshalingError defines Unmarshaling/Marshaling errors codes,
```
trent-reed, line 87: Needs to be a docstring comment.

errors/src/lib.rs
```diff
@@ -37,3 +39,57 @@ impl TryFrom<u32> for TssError {
...
+// providing more explicit error codes for try_marshal* and try_unmarchal*.
+#[derive(PartialEq, Eq, Copy, Clone, Debug)]
+pub enum MarshalingError {
+    ArrayLengthExceeded,
+    UnexpectedEndOfBuffer,
+    UnknownSelector,
+    MarshalingDeriveError,
```
trent-reed, line 94: Nit: Maybe sort alphabetically.

errors/src/lib.rs
```diff
@@ -37,3 +39,57 @@ impl TryFrom<u32> for TssError {
...
+            Self::ArrayLengthExceeded => {
+                write!(f, "length of array is larger than the type allows")
+            }
+            Self::UnexpectedEndOfBuffer => {
+                write!(f, "expected to have more buffer data but found none")
+            }
+            Self::UnknownSelector => {
+                write!(f, "selector targeted is not known to the unmarshaling code")
+            }
+            Self::MarshalingDeriveError => write!(f, "unexpected derive error"),
```
trent-reed, line 60: Nit: Sort match alphabetically

errors/src/lib.rs
```diff
@@ -37,3 +39,57 @@ impl TryFrom<u32> for TssError {
...
+    }
+}
+
+impl From<MarshalingError> for TpmRcError {
+    fn from(orig: MarshalingError) -> Self {
+        match orig {
+            MarshalingError::ArrayLengthExceeded => TpmRcError::Size,
+            MarshalingError::UnexpectedEndOfBuffer => TpmRcError::Memory,
+            MarshalingError::UnknownSelector => TpmRcError::Selector,
+            MarshalingError::MarshalingDeriveError => TpmRcError::Failure,
```
trent-reed, line 71: Nit: Sort match alphabetically

errors/src/lib.rs
```diff
@@ -37,3 +39,57 @@ impl TryFrom<u32> for TssError {
...
+    }
+}
+
+impl From<MarshalingError> for TssError {
+    fn from(orig: MarshalingError) -> Self {
+        match orig {
+            MarshalingError::ArrayLengthExceeded => TpmRcError::Size.into(),
+            MarshalingError::UnexpectedEndOfBuffer => TpmRcError::Memory.into(),
+            MarshalingError::UnknownSelector => TpmRcError::Selector.into(),
+            MarshalingError::MarshalingDeriveError => TpmRcError::Failure.into(),
```
trent-reed, line 82: Nit: Sort match alphabetically

base/src/commands/capability.rs
```diff
@@ -2,7 +2,7 @@
 
 use crate::commands::{Marshalable, TpmCommand};
 use crate::constants::{TpmCap, TpmCc, TpmPt};
-use crate::{TpmiYesNo, TpmsCapabilityData};
+use crate::{MarshalingResult, TpmiYesNo, TpmsCapabilityData};
```
trent-reed, line 5: Wait wait, we should not need to add these to these files... If we're doing that a derive macro is wrong somewhere. Here, and other files where you aren't explicitly referencing `MarshalingResult`, please remove these added using statements.

marshalable-derive/src/lib.rs
```diff
@@ -186,11 +186,11 @@ fn derive_tpm_marshal_inner(input: DeriveInput) -> Result<TokenStream> {
             #pure_impl
             // The generated impl.
             impl Marshalable for #name  {
-                fn try_unmarshal(buffer: &mut tpm2_rs_marshalable::UnmarshalBuf) -> tpm2_rs_marshalable::exports::errors::TpmRcResult<Self> {
+                fn try_unmarshal(buffer: &mut tpm2_rs_marshalable::UnmarshalBuf) -> MarshalingResult<Self> {
```
trent-reed, line 189: You need to fully name this (`tpm2_rs_marshalable::exports::errors::MarshalingResult`, or whatever the final path will be). Here and elsewhere for the derive macro.

This is why you needed to add use statements everywhere.

marshalable/src/lib.rs
```diff
@@ -31,10 +31,10 @@ pub mod exports {
 // entries that should be marshaled. See TpmlPcrSelection for an example.
 pub trait Marshalable: Sized {
     // Unmarshals self from the prefix of `buffer`. Returns the unmarshalled self and number of bytes used.
-    fn try_unmarshal(buffer: &mut UnmarshalBuf) -> TpmRcResult<Self>;
+    fn try_unmarshal(buffer: &mut UnmarshalBuf) -> MarshalingResult<Self>;
```
trent-reed, line 34: Note: In this file you ***DON'T*** need to fully-qualify these names, since these aren't part of the derive macro. No action needed here.

marshalable/src/tests.rs
```diff
@@ -11,7 +11,7 @@ macro_rules! impl_test_scalar {
         let same_size_buffer: [u8; SIZE_OF_TYPE] = [$I; SIZE_OF_TYPE];
         let larger_buffer: [u8; SIZE_OF_TYPE + 4] = [$I; SIZE_OF_TYPE + 4];
 
-        let mut res: TpmRcResult<$T> =
+        let mut res: MarshalingResult<$T> =
```
trent-reed, line 14: Nit: Tad strange that we're explicitly listing the type here, do we need to do this? Can we just `let mut res = <$T>::try_unmarshal(&mut UnmarshalBuf::new(&too_small_buffer));`?

This whole thing is a tad funny of test setup actually, so maybe this is just a nit since you didn't add this. We can probably clean it up better in a separate PR.

server/src/tests/drbg.rs
```diff
@@ -10,7 +10,7 @@ pub struct FakeDrbgError;
 
 impl Display for FakeDrbgError {
     fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
-        write!(f, "{:#?}", self)
+        write!(f, "{self:#?}")
```
trent-reed, line 13: Thanks for the cleanup on this (here and elsewhere)!

trent-reed (commented):
I think we need some other eyes on this to make sure everyone agrees on the right direction, I added some clarifying comments and fix suggestions. Let's add @josephlr at least, probably @oddcoder as well (it says you're busy oddcoder, so if you do not have time just let us know).

oddcoder:
> probably @oddcoder <https://github.com/oddcoder> as well (it says you're
busy oddcoder, so if you do not have time just let us know)


Hi, appologies for late responses, I am a bit on tight schedule, feel free
to align without me.

Best,

base/src/commands/capability.rs
```diff
@@ -1,7 +1,7 @@
 //! [TPM2.0 1.83] 30 Capability Commands
-
 use crate::commands::{Marshalable, TpmCommand};
 use crate::constants::{TpmCap, TpmCc, TpmPt};
+use crate::marshal;
```
trent-reed, line 4: This still should not be needed if you are setting up the proc macro properly. There should be no changes at all to these files that simply use the proc-macro if things are done correctly. I'll comment below for where the issue is if I find it.

marshalable-derive/src/lib.rs
```diff
@@ -92,13 +92,13 @@ fn derive_tpm2b_struct(
         impl Tpm2bStruct for #tpm2b_outer_struct_name {
             type StructType = #tpm2b_inner_struct_name;
 
-            fn from_struct(val: &Self::StructType) -> TpmRcResult<Self> {
+            fn from_struct(val: &Self::StructType) -> marshal::Result<Self> {
```
trent-reed, line 95: Here (and below), you must fully-qualify the type. The original code erroneously did not do this.

So here, this should be `tpm_rs_marshalable::Result`, then in the future when we move those types to `tpm_rs_base` it'll turn into something like `tpm_rs_base::marshal::Result`. The point is you shouldn't need to add any includes other than the proc-macro at the callsite.

marshalable/src/errors.rs
```diff
@@ -0,0 +1,57 @@
+use core::error::Error;
+use core::fmt;
+use tpm2_rs_errors::*;
+
+/// Represents success or [`MarshalingError`] failure, which is used for unmarshal/marshal functionality.
+pub type MarshalingResult<T> = Result<T, MarshalingError>;
```
trent-reed, line 6: If we're going to do this just to rename it in the base module, just name it `Result` to begin with, and then don't rename it in the base module.

If we find the shared name to be confusing, we can just not have a type named `Result` and at all the locations where there's a `Result` fill in the expected error type (in this case `MarshalingError`). I think in the end @josephlr should be the one to make that call, but for now let's just get something submitted, and then we can clean it up later.

marshalable/src/errors.rs
```diff
@@ -0,0 +1,57 @@
...
+
+/// The MarshalingError defines Unmarshaling/Marshaling errors codes,
+/// providing more explicit error codes for try_marshal* and try_unmarshal*.
+#[derive(PartialEq, Eq, Copy, Clone, Debug)]
+pub enum MarshalingError {
+    ArrayLengthExceeded,
+    MarshalingDeriveError,
```
trent-reed, line 13: If there's an error during `derive` it's a compiler error, not a runtime error. I highly suspect you can just delete this enum variant. This seems unneeded to me.

marshalable/src/errors.rs
```diff
@@ -0,0 +1,57 @@
...
+use core::fmt;
+use tpm2_rs_errors::*;
+
+/// Represents success or [`MarshalingError`] failure, which is used for unmarshal/marshal functionality.
+pub type MarshalingResult<T> = Result<T, MarshalingError>;
+
+/// The MarshalingError defines Unmarshaling/Marshaling errors codes,
+/// providing more explicit error codes for try_marshal* and try_unmarshal*.
+#[derive(PartialEq, Eq, Copy, Clone, Debug)]
+pub enum MarshalingError {
```
trent-reed, line 11: Ditto here, if we're just going to rename this to `Error` then just name it `Error` to begin with and don't add a `use core::error::Error;` to this file.

marshalable/src/lib.rs
```diff
@@ -6,14 +6,13 @@ use core::option::{Option, Option::*};
...
 pub use tpm2_rs_marshalable_derive::Marshalable;
 pub use tpm2_rs_marshalable_derive::Tpm2bStruct;
 
-/// Exports needed for macro expansion
-pub mod exports {
-    pub use tpm2_rs_errors as errors;
-}
+/// Expose the submodule & use the marshalling error.
+mod errors;
+pub use errors::MarshalingError as Error;
```
trent-reed, line 14: I'd instead just name these appropriately in `errors`, and then here we can do `mod errors; pub use errors::*;` to import all the publics into this module without needing to rename them manually.

marshalable/src/errors.rs
```diff
@@ -0,0 +1,57 @@
+use core::error::Error;
+use core::fmt;
+use tpm2_rs_errors::*;
+
+/// Represents success or [`MarshalingError`] failure, which is used for unmarshal/marshal functionality.
+pub type MarshalingResult<T> = Result<T, MarshalingError>;
+
+/// The MarshalingError defines Unmarshaling/Marshaling errors codes,
+/// providing more explicit error codes for try_marshal* and try_unmarshal*.
+#[derive(PartialEq, Eq, Copy, Clone, Debug)]
```
trent-reed, line 10: Personally I would also add `Hash, Ord, PartialOrd` as well, I usually do so for all simple non-data-carrying enum types.

marshalable/src/errors.rs
```diff
@@ -0,0 +1,57 @@
...
+            MarshalingError::MarshalingDeriveError => TpmRcError::Failure,
+            MarshalingError::UnexpectedEndOfBuffer => TpmRcError::Memory,
+            MarshalingError::UnknownSelector => TpmRcError::Selector,
+        }
+    }
+}
+
+impl From<MarshalingError> for TssError {
+    fn from(orig: MarshalingError) -> Self {
+        match orig {
```
trent-reed, line 50: This whole match block could be replaced with just a simple: `TpmRcError::from(orig).into()`

Not that it matters much, as I've said in the past eventually I think `TssError` should go away. But might as well be concise while we have it around.

marshalable/src/errors.rs
```diff
@@ -0,0 +1,57 @@
...
+                write!(f, "expected to have more buffer data but found none")
+            }
+            Self::UnknownSelector => {
+                write!(f, "selector targeted is not known to the unmarshaling code")
+            }
+        }
+    }
+}
+
+impl From<MarshalingError> for TpmRcError {
```
trent-reed, line 37: Note: Eventually, this might make more sense to put in the same file as `TpmRcError`, but I think because of the funky crate layout we have today, it has to go here. After the crates are simplified down, this probably will be able to go where it belongs.

Issue #171, opened 2025-06-23
Document how to refresh CLASIG_PAT


chrisfenner:
It seems CLASIG_PAT has expired (example failure: https://github.com/tpm-rs/tpm-rs/actions/runs/15704497924).

Could we refresh this access token? I'm not sure how it was created, so maybe now is a good time to document how it was created.

josephlr:
I'm working on this right now. I think we can also setup a token that just never expires.

PR #194, opened 2025-08-15
Update cla.yml to fix CLA Assistant

josephlr:
This is part of addressing https://github.com/tpm-rs/tpm-rs/issues/171


josephlr:
Bypassing approval here so that I can actually get the CLA Check to pass (it runs on the base branch).

PR #200, opened 2025-11-03
tpm2: rework of base library crate

josephlr:
This PR contains a prototype of how we might handle types, marshalling, and unmarshalling for `tpm2`.

This is mostly just to get feedback at the TPM-RS Alignment meeting today.

The essential problem addressed here is to improve:
  - Stack Size of TPM2 data structures
  - Code Size of Marshalling/Unmarshalling code

while still having a _single_ version of all of our types.

TSS libraries (in Rust or C) often face a choice between:
  - Option 1: Use a large number of `#define`s and conditional compilation to compile exactly the algorithms and buffer sizes they support and no more.
    - Pro: Stack size of structures is smaller (for example, if you only want to support 32-byte hash algorithms, your `TPMT_HA`, `TPMU_HA`, `TPM2B_DIGEST`, etc... will all be half the size).
    - Pro: Marshalling/Unmarshalling code size is smaller. You don't need to generate code for reading/writing unused variants
    - Con: Requires a lot of conditional compilation see [the `TpmConfiguration` folder in the reference code](https://github.com/TrustedComputingGroup/TPM/tree/main/TPMCmd/TpmConfiguration/TpmConfiguration)
    - Con: the types are now different (in both API and ABI) and cannot be shared between users of a common library
  - Option 2: Simply support ~all algorithms and make an assumption about what a "reasonable" maximum buffer lengths should be for various `TPM2B_` types.
    - Pro: Simpler
    - Pro: Stable API / ABI, this is what [tpm2-tss does](https://github.com/tpm2-software/tpm2-tss/blob/2270f1e8fd1cb93a57c98eb8b03f2067bacdad03/include/tss2/tss2_tpm2_types.h#L22-L44)
    - Con: Size of structures increases if you want to support most algorithms
    - Con: Code size of marshalling increases to support the additional variants
    - Con: What if your "reasonable" assumptions are wrong? For example, if a TPM2 has `TPM2_MAX_CAP_BUFFER > 1024`, the TPM2-TSS will be unable to correctly parse some responses from `GetCapability`.

My idea is to solve these issues in the Rust code in two ways:

1.   Improve the stack-size situation. by storing references to buffers/arrays in the TPM2 structures instead of storing them by values. For example, if we have a `TpmtHa` which looks like:
```rust
pub enum TpmtHa<'a> {
    Sha1(&'a [u8; 20]),
    Sha256(&'a [u8; 32]),
    Sha384(&'a [u8; 48]),
    Sha512(&'a [u8; 64]),
}
```
then `TpmtHa` will always have the size of two pointers, regardless of how many additional variants we add. It also improves unmarshalling as we no longer need to `memcpy` over the bytes.

2. Introduce a `Limits` trait, which says "hey I support this algorithm" and "this my `TPM2_MAX_CAP_BUFFER` value". Then our `Marshal` and `Unmarshal` traits can take a `L: Limit` generic parameter, which is used to determine if we should try marshalling or unmarshalling certain variants. This allows for compile-time determination of what limits are reasonable. This allows for the code size savings while still having a single `TpmtHa` type.

I was able to confirm things work as expected with the sample code:
```rust
use tpm2::errors::{MarshalError, UnmarshalError};
use tpm2::marshal::{EccCurves, Limits, Marshal, RsaKeySizes, TpmaHashAlgs, Unmarshal};
use tpm2::{TpmiAlgHash::*, TpmiEccCurve, TpmtHa};

pub struct Small;
impl Limits for Small {
    const HASH_ALGS: TpmaHashAlgs = TpmaHashAlgs::from_alg_list(&[Sha256]);
    const RSA_KEY_SIZES: RsaKeySizes = RsaKeySizes::NONE;
    const ECC_CURVES: EccCurves = EccCurves::from_curve_list(&[TpmiEccCurve::NIST_P256]);
}

pub struct Medium;
impl Limits for Medium {
    const HASH_ALGS: TpmaHashAlgs = TpmaHashAlgs::from_alg_list(&[Sha1, Sha256]);
    const RSA_KEY_SIZES: RsaKeySizes = RsaKeySizes::from_key_bits_list(&[2048]);
    const ECC_CURVES: EccCurves = EccCurves::from_curve_list(&[
        TpmiEccCurve::NIST_P256,
        TpmiEccCurve::NIST_P384,
        TpmiEccCurve::BN_P256,
    ]);
}

pub struct Large;
impl Limits for Large {
    const HASH_ALGS: TpmaHashAlgs = TpmaHashAlgs(u32::MAX);
    const RSA_KEY_SIZES: RsaKeySizes = RsaKeySizes::ALL;
    const ECC_CURVES: EccCurves = EccCurves::ALL;
}

pub fn marshal_small(ha: TpmtHa, buf: &mut [u8]) -> Result<(), MarshalError> {
    ha.marshal::<Small>(buf)?;
    Ok(())
}
pub fn marshal_medium(ha: TpmtHa, buf: &mut [u8]) -> Result<(), MarshalError> {
    ha.marshal::<Medium>(buf)?;
    Ok(())
}
pub fn marshal_large(ha: TpmtHa, buf: &mut [u8]) -> Result<(), MarshalError> {
    ha.marshal::<Large>(buf)?;
    Ok(())
}

pub fn unmarshal_small(buf: &[u8]) -> Result<TpmtHa<'_>, UnmarshalError> {
    let mut ha = TpmtHa::Sha256(&[0; Sha256.digest_size()]);
    ha.unmarshal::<Small>(buf)?;
    Ok(ha)
}
pub fn unmarshal_medium(buf: &[u8]) -> Result<TpmtHa<'_>, UnmarshalError> {
    let mut ha = TpmtHa::Sha256(&[0; Sha256.digest_size()]);
    ha.unmarshal::<Medium>(buf)?;
    Ok(ha)
}
pub fn unmarshal_large(buf: &[u8]) -> Result<TpmtHa<'_>, UnmarshalError> {
    let mut ha = TpmtHa::Sha256(&[0; Sha256.digest_size()]);
    ha.unmarshal::<Large>(buf)?;
    Ok(ha)
}
```

And I got the following code size results with `opt-level = 2` (works better than `opt-level = "s"`):

Function | `riscv32imc-unknown-none-elf` | `thumbv7m-none-eabi`
--------- | ------------ | ---------
`code_size::marshal_large` | 104 bytes | 190 bytes 
`code_size::marshal_medium` | 85 bytes | 134 bytes 
`code_size::marshal_small` | 60 bytes | 62 bytes 
`code_size::unmarshal_large` | 99 bytes | 140 bytes 
`code_size::unmarshal_medium` | 84 bytes | 76 bytes 
`code_size::unmarshal_small` | 52 bytes | 54 bytes


josephlr:
Closing in favor of #247

PR #213, opened 2026-02-13
Allow for the creation of multiple TcpSimulators.

nnooney:
- The TCG reference simulator writes the ports it uses to `command.port` and `platform.port` in the directory of the spawned simulator process. Rather than assume default ports, the TcpConnection associated with the simulator now parses those files for the ports to use.
- Combine that with the use of a temporary directory for each integration test and now the integration tests all run in parallel (we can drop the `--test-threads=1` parameter from the tests).

Before: `test result: ok. 5 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 5.03s`
After: `test result: ok. 5 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 1.01s`

- Since the simulator now runs in parallel, also provide utilities for capturing stdout/stderr. This prevents the code from interleaving test output. As part of this change, we no longer need the `nocapture` flag, since this feature now allows the capture and handling of the simulator. It's particularly useful in tests like this:

```rust
#[test]
simulator_test() -> Result<()> {
    let mut simulator = TcpSimulator::new( ... )?;

    // run command that returns result ...
    let result = ...

    if result.is_err() {
        if simulator.stop_nicely().is_ok()
            && let Ok(output) = simulator.collect_output()
        {
            println!("{output}");
        }
    }
}
```

Future improvements:

- Don't persist the temporary directory via the `keep()` function. This isn't a big deal since we're running in an ephemeral container anyways, but it's possible to refactor the test code to properly maintain the scope of the temporary directory.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,6 +20,7 @@ proc-macro2 = "1"
 quote = "1"
 safe-discriminant = "0.2.0"
 syn = { version = "2", features = ["full"] }
+tempfile = "3.25.0"
 trybuild = { version = "1.0.89", features = ["diff"] }
 zerocopy = { version = "0.8.33", features = ["derive"] }
 
--- a/client/Cargo.toml
+++ b/client/Cargo.toml
@@ -16,9 +16,10 @@ zerocopy = { workspace = true, optional = true }
 
 [dev-dependencies]
 tpm2-rs-unionify = { workspace = true }
+tempfile = { workspace = true }
 
 [[test]]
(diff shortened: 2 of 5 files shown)


josephlr (changes requested):
I like this; thanks for working on this and sorry for the review dealy!

Attached comments from our discussion during the alignment meeting.

josephlr (approved):
Looks good! Thanks for getting this in!

PR #216, opened 2026-04-22
Set package keys in workspace file and update to Rust 2024

josephlr:
Also run `cargo fmt` repo-wide.


josephlr:
Merging to unblock folks (ideally would be reviewed later once people are back in office).

PR #223, opened 2026-05-27
Integrate go-tpm tests via TCP connection

crocomegan:
Introduced a new integration test run_simulator_for_go_tests in crates/tpm2-client/tests/simulator.rs that fires up the simulator, drops port configurations into go_tests/command.port and go_tests/platform.port, and invokes go test automatically.


crates/tpm2-client/src/connection/tcp.rs
```diff
@@ -713,6 +723,16 @@ impl TcpSimulator {
     pub fn connection_mut(&mut self) -> &mut TcpConnection {
         &mut self.conn
     }
+
+    /// Returns the TPM port of the running simulator.
+    pub fn tpm_port(&self) -> u16 {
```
nnooney, line 728: The TCP Simulator reads the tpm_port and plat_port when it's started (around line 673 above). Why not save the ports as member variables on the TcpSimulator struct and reference them directly?

crates/tpm2-client/Dockerfile
```diff
@@ -85,3 +85,13 @@ WORKDIR /tpm-rs
 RUN curl https://sh.rustup.rs -sSf | bash -s -- -y
 ENV PATH="/root/.cargo/bin:${PATH}"
 RUN cargo --version
+
+# Install Go 1.24
```
nnooney, line 89: Any reason not to use the Ubuntu package repository which has go 1.24?

https://documentation.ubuntu.com/ubuntu-for-developers/reference/availability/go/#ubuntu-golang-deb-packages

`golang-1.24` is the package name: https://launchpad.net/ubuntu/+source/golang-1.24

crates/tpm2-client/compose.yaml
```diff
@@ -3,4 +3,5 @@ services:
     build: .
     volumes:
       - ../../:/tpm-rs
+      - ../../../go-tpm:/go-tpm
```
nnooney, line 6: This assumes go-tpm is installed at a particular location on the host system (namely in the same directory as tpm-rs) and will cause failures if the volume doesn't exist. Instead of mounting it in the main compose.yaml file, I think we should use the extends feature of Docker to optionally refer to the go-tpm volume.

https://docs.docker.com/compose/how-tos/multiple-compose-files/extends/

e.g. there would be a 'go-tpm.yaml' Docker compose file that extends this 'compose.yaml' file and you'd run the go-tpm mounted variant with `docker compose run --rm -f go-tpm.yaml simulator_tests`.

crates/tpm2-client/tests/simulator.rs
```diff
@@ -68,3 +68,28 @@ fn get_started_tpm() -> TcpSimulator {
     run_command(&startup, simulator.connection_mut()).unwrap();
     simulator
 }
+
+#[test]
+fn run_simulator_for_go_tests() {
```
nnooney, line 73: I'm curious what the benefit of launching the Go tests from within Cargo is as opposed to having tpm-rs expose a simulator binary build target that we can use directly with go-tpm.

It seems to me that we're mixing the current Docker container used to build the TCG reference TPM Simulator with a Docker container to run go-tests using a go toolchain.

PR #248, opened 2026-07-16
Add v185 specs to docs

josephlr:
Having a PDF copy in-tree should make it easier to reference (both for humans and AI tools)


nnooney (approved):
Let's also add a README.md to the tpm2_v185 folder that contains the canonical URL where these docs came from.

josephlr:
> Let's also add a README.md to the tpm2_v185 folder that contains the canonical URL where these docs came from.

Done!

Issue #243, opened 2026-07-13
Release 0.1 of tpm2-base and tpm2-client


nnooney:
This bug tracks the steps necessary to release version 0.1 of the following crates:

- `tpm2`, from `crates/tpm2`.
- `tpm2-client`, from `crates/tpm2-client`.

No backwards-compatibilities are guaranteed with this release.

# Checklist

- [ ] Blocking Issues closed (see `Blocked Issues` below)
- [ ] Rust API Guidelines Check: https://rust-lang.github.io/api-guidelines/checklist.html
- [ ] RustDoc Review
- [ ] GitHub Actions Integration
- [ ] Creating 0.1 release in GitHub

# Blocked Issues

- [ ] #215 
- [x] #244

PR #249, opened 2026-07-20
Make types eagerly implement common traits

nnooney:
Per Rust API Guidelines, types should eagerly implement common traits.

https://rust-lang.github.io/api-guidelines/interoperability.html#types-eagerly-implement-common-traits-c-common-traits


nnooney:
#243 blocker

josephlr (approved):
Debug, Clone, Copy, PartialEq, Eq, and Hash all seem reasonable for all structs to implement.

PR #252, opened 2026-08-01
Fix API naming violations and getter conventions in tpm2 and tpm2-cli…

nnooney:
…ent crates.

- Renamed TpmRcError constructors (AsymmetricFor -> asymmetric_for, ValueFor -> value_for, SizeFor -> size_for, SelectorFor -> selector_for).
- Fixed typo in KeyAlreadyRegistered constant name in tss_rc.
- Renamed get_auth_command to auth_command across Session trait and implementations.
- Renamed PasswordSession::get_secret to PasswordSession::secret.
- Renamed TcpConnection getters (get_command_response_sizes -> command_response_sizes, act_get_signaled -> act_signaled).
- Fixed boolean logic and parameter semantics in CmdHeader::new.

PR #253, opened 2026-08-01
Implement standard error and interoperability traits across tpm2 and tpm2-client.

nnooney:
- Implemented core::fmt::Display and core::error::Error for MarshalError, UnmarshalError, HashError, TssErrorCannotBeZero, TpmRcError, and generated TSS layer errors.
- Derived standard traits (Hash, Default, Eq, Clone) on Alg, TpmiAlgHash, TpmtHa, GetRandom, CmdHeader, RespHeader, PasswordSession, and TCP connection structs/enums.

### DIFF
--- a/crates/tpm2-client/src/connection/tcp.rs
+++ b/crates/tpm2-client/src/connection/tcp.rs
@@ -232,7 +232,7 @@ impl TcpConnection {
     ///
     /// Returns an error if the communication with the Platform port fails or if
     /// the response terminator is invalid.
-    pub fn get_command_response_sizes(&mut self) -> Result<GetCommandResponseSizesResponse> {
+    pub fn command_response_sizes(&mut self) -> Result<GetCommandResponseSizesResponse> {
         let cmd_code = U32::new(SimulatorPlatformCommandCode::GetCommandResponseSizes as u32);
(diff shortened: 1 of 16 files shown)


nnooney:
Don't review this until #252 is reviewed / submitted.

#243 

I think I need to update my workflow now that stacked PRs are live: https://docs.github.com/en/pull-requests/get-started/about-stacked-prs

> Stacked pull requests require all branches to be in the same repository. Cross-fork stacks are not supported.

josephlr (changes requested):
I agree on the `PartialEq, Eq` implementations, and the `Debug, Error, Display` implementations for the error types.

I'm not sure we should implement `Default` this widely, as not all enums have a "reasonable" default, and because of that I don't think we can implement `Default` for all the `Commands`/`Responses`, but I might be wrong there.

As for `Hash`, my only concern there is that we don't really use it, so it might hurt compile times due to macro-expansion and code-gen. Thoughts?

nnooney:
I'll update this:

- No hash implementations (YAGNI), can add later when proven to be needed.
- Only implement `Debug`, `Error`, and `Display` for error types.
- Remove `Default` (don't pick default enumerations)

phil-levis-google:
Any update?

PR #256, opened 2026-08-13
Remove base / marshalable / marshalable-derive, and implement the remaining tpm2 structures

josephlr:
## Summary

Refactors the workspace to consolidate core TPM 2.0 primitives into the `tpm2` crate.

## Key Changes
- **Workspace Cleanup**: Deleted `base`, `marshalable`, and `marshalable-derive` crates; cleaned root `Cargo.toml`.
- **Part 2 Primitives**: Added complete tables of TPM 2.0 constants, error types (`TpmRc`), Part 2 structures (`TPM2B`, `TPMA`, `TPMI`, `TPMS`, `TPML`, `TPMU`, `TPMT`), and standard headers (`CommandHeader`, `ResponseHeader`).
- **Marshalling Architecture**:  Used the pre-existing `tpm2::Marshal` and `tpm2::Unmarshal<'a>` traits with fixed `MAX_SIZE` bounds, replacing proc-macro derive serialization.
  - For now, all `Marshal`/`Unmarshal` implementations are manual (no `proc_macro` or `macro_rules`).
- Moved `Connection` to `tpm2`
- Added initial command definitions (`GetRandom`, `StirRandom`, `Startup`, `Shutdown`).
- **Error Types**
  - **Structured `TpmRc`**: Replaced legacy `TpmRcError` enum hierarchy with a consolidated 32-bit `TpmRc` struct supporting Format 0, Format 1 (parameter, handle, session modifiers), and vendor-specific response codes.
  - **`ResponseHeader` Integration**: `ResponseHeader.rc` directly encodes success/failure as `Result<(), TpmRc>`.
  - **Legacy Removal**: Deleted `TssError`, `TssRcError`, `TssTcsError`, and `TssTctiError` abstractions.
- **Crate Migrations**:
  - `tpm2-impl`: Switched from `tpm2-rs-base` to `tpm2` constants and `TpmRc` errors.
  - `tpm2-client`: Updated command execution (`run_command`, `run_command_with_sessions`) and session handling to use refactored `tpm2` structures.

## Verification
- `cargo check`: Passed (0 errors, 0 warnings)
- `cargo fmt --check`: Passed
- `cargo clippy -- -D warnings`: Passed (0 warnings)
- `cargo test`: Passed (62 unit/integration/doc tests)


crates/tpm2/src/lib.rs
```diff
@@ -57,20 +57,50 @@
 //! Internally, we use submodules for code organization, but mostly present a
 //! flat API to external users, with the exception of the [`commands`] and
 //! [`errors`] submodules.
-#![no_std]
+#![cfg_attr(not(any(feature = "std", test)), no_std)]
```
wenyuzhao: Add #![no_std] to all top-level modules other than std.rs? This ensures all other files do not contain std code even if "std" feature is enabled.

crates/tpm2-client/src/sessions/session.rs
```diff
@@ -1,9 +1,26 @@
-use tpm2_rs_base::{TpmsAuthCommand, TpmsAuthResponse, errors::TssResult};
+use tpm2::{TpmsAuthCommand, TpmsAuthResponse};
+
+/// Error occurring during session authorization validation.
+#[derive(Debug, PartialEq, Eq, Clone, Copy)]
+pub enum AuthError {
```
wenyuzhao, line 5: Should this and all session-related code be moved to tpm2? Since TpmsAuthCommand, TpmsAuthResponse is in tpm2 already.

crates/tpm2/src/structures/tpmi.rs
```diff
@@ -0,0 +1,308 @@
...
+        }
+    }
+}
+
+/// `TPMI_ALG_HASH` interface type defined in TPM 2.0 Part 2: Structures, Section 9.21 (Table 52).
+///
+/// Selects a hash algorithm (SHA1, SHA256, SHA384, SHA512, SM3_256, etc.).
+/// Note: `TPM_ALG_NULL` is represented as `Option<TpmiAlgHash>::None`.
+#[derive(Copy, Clone, PartialEq, Eq, Debug)]
+#[repr(u16)]
```
wenyuzhao: This can probably be a `u8`? same as other enums like TpmiStCommandTag

crates/tpm2/src/errors/tpm_rc.rs
```diff
@@ -0,0 +1,462 @@
...
+    }
+}
+
+#[derive(PartialEq, Eq, Clone, Copy)]
+pub struct Position(NonZero<u32>);
+
+impl Position {
+    /// Position for a handle (1-based index, 1..=7).
+    #[inline]
+    pub const fn handle(n: u8) -> Self {
```
wenyuzhao, line 397: We may need to review all the uses of `#[inline]` attributes, to control code size.

wenyuzhao (approved):
LGTM

crates/tpm2-client/src/lib.rs
```diff
@@ -27,94 +27,201 @@
...
-        .map_err(TssError::from)?;
-
-    written += cmd_handles
-        .try_marshal(&mut cmd_buffer[written..])
-        .map_err(TssError::from)?;
+    let mut cmd_header = CommandHeader::with_sessions(!cmd_sessions.is_empty(), CmdT::CMD_CODE);
+    let mut written = cmd_header.marshal(
+        (&mut cmd_buffer[0..CommandHeader::MAX_SIZE])
+            .try_into()
+            .unwrap(),
```
nnooney, line 181: I know we wanted to enforce no panics; do we need unwrap here?

nnooney (approved):

crates/tpm2/src/structures/tpmu.rs
```diff
@@ -0,0 +1,190 @@
...
+
+/// Internal union representing public object parameters and unique identifier.
+#[derive(Clone, Copy, PartialEq, Eq, Debug)]
+#[repr(u16)]
+pub enum PublicParmsAndId {
+    KeyedHash(Option<TpmtKeyedHashScheme>, Tpm2bDigest) = Alg::KEYEDHASH.tag(),
+    Sym(TpmtSymDefObject, Tpm2bDigest) = Alg::SYMCIPHER.tag(),
+    Rsa(TpmsRsaParms, Tpm2bPublicKeyRsa) = Alg::RSA.tag(),
+    Ecc(TpmsEccParms, TpmsEccPoint) = Alg::ECC.tag(),
+}
```
josephlr: After talking with @nnooney we should:
- Add a bunch more comments about this
- move it to `structures/mod.rs` as it isn't a direct mapping to a single `tpmu` type, but rather a pair of them.

crates/tpm2/src/structures/tpmi.rs
```diff
@@ -0,0 +1,308 @@
...
+        Alg::unmarshal(src)?.try_into()
+    }
+}
+
+/// `TPMI_RSA_KEY_BITS` interface type defined in TPM 2.0 Part 2: Structures, Section 9.29 (Table 60).
+///
+/// Selects an RSA key size in bits (1024, 2048, 3072, or 4096 bits).
+#[derive(Clone, Copy, PartialEq, Eq, Debug)]
+#[repr(transparent)]
+pub struct TpmiRsaKeyBits(pub u16);
```
josephlr: After discussing with @nnooney, we may want to do something like:
```rust
pub enum TpmiRsaKeyBits {
    Rsa1024 = 1024,
    Rsa2048 = 2048,
    Rsa3072 = 3072,
    Rsa4096 = 4096,
}
```

So that there's an explicit set of key sizes we can toggle with cargo features (in a future PR).

josephlr: Done

crates/tpm2/src/structures/tpmu.rs
josephlr: Done

crates/tpm2-client/src/lib.rs
josephlr: Given that `CommandHeader::MAX_SIZE == 10 < CMD_BUFFER_SIZE = 0x1000` this should never panic.

We use a similar style in `tpm2::marshal::marshal_helper` to avoid panics with consise code.

crates/tpm2/src/errors/tpm_rc.rs
josephlr: All inline uses removed

crates/tpm2/src/structures/tpmi.rs
josephlr: all `repr(...)` tags on the enums are now removed, so the compiler can determine the size.

crates/tpm2/src/lib.rs
josephlr: Good point, changed to `#![cfg_attr(not(test), no_std)]` to prevent this.

crates/tpm2-client/src/sessions/session.rs
josephlr: I don't think so right now. session authorization validation is tricky, and the checks are going to be different client-side vs impl-side. If in the future there are some common checks, we should move those to `tpm2`, but we can handle that in a followup PR.

josephlr:
This PR is ready for review at the Alignment meeting.

Some specific things for folks to look at:
- Definitions of `Marshal` and `Unmarshal`
- How a "typical" `Tpms*` or `Tpmt**` type works
- How some "exceptional" types work:
  - `TpmtPublic` + `PublicParmsAndId`
  - `TpmsAttest` + `TpmGenerated` + `TpmuAttest`
  - `TpmtSensitive` + `TpmuSensitiveComposite`
  - `TpmsPcrSelect`
  
Some things which need further review or followup PRs:
- `Tpml*` types
  - Right now they are not generic over length and element type
  - They currently require an underlying element type which implements `Default`
  - This can be an issue for types like `TpmlPcrSelection` with elements `TpmsPcrSelection`
  - We can cleanup the Marshal/Unmarshal implementations if they are generic.
  - The max capacity of these `Tpml*` types is sometimes determined by the spec, but is sometimes implementation-defined, how should we handle this?
- `Tpm2b*` types
  - Right now they are not generic over length
  - Right now they do not take the slice by reference but instead copy it into a statically-sized buffer.
  - We can cleanup the Marshal/Unmarshal implementations if they are generic.
  - The max capacity of these`Tpm2b*` types is sometimes determined by the spec, but is sometimes implementation-defined, how should we handle this?
  
Do we want to fix these things before or after merging this PR?

nnooney (approved):

phil-levis-google:
Comment from Benjamin Doron during review today in the public meeting: add comments explaining the const vs. type structure of tyoes like Alg; describe how this is a general pattern.

PR #261, opened 2026-09-03
Crypto Interfaces

josephlr:
This is a draft of how our Crypto interfaces might look.
- A backend defines a `Hash` implementation, which specifies things like:
  - The `Hash::Sha256Ctx` associated type
  - Implementations for:
    - `Hash::sha256()` constructing a `Sha256Ctx`
    - `Sha256Ctx::update()`
    - `Sha256Ctx::finalize()`
  - Uses code like:
    ```rust
    type Sha3_512Ctx = !;
    ```
    to indicate a backend doesn't support `Sha3_512`.
- We then define helper types / functions to handle things like:
  - Init/Update/Finalize hash APIs via `HashCtx`'s `new`/`update`/`finalize` methods.
  - One-shot hashing via:
    ```rust
    pub fn hash<'a, H: Hash>(
        h: &H,
        alg: TpmiAlgHash,
        data: &[u8],
        out: &'a mut [u8; TpmtHa::MAX_DIGEST_SIZE],
    ) -> Result<TpmtHa<'a>, H::Error>
    ```

### DIFF
--- a/crates/tpm2/Cargo.toml
+++ b/crates/tpm2/Cargo.toml
@@ -16,6 +16,16 @@ categories.workspace = true
 # Implement conversions/traits which require std
 std = []
 
+# Only enable SHA2 by default
+default = ["sha256", "sha384", "sha512"]
+sha1 = []
+sha256 = []
+sha384 = []
+sha512 = []
+sm3_256 = []
+sha3_256 = []
+sha3_384 = []
+sha3_512 = []
+
 [dependencies]
 bitflags = { workspace = true }
-
--- a/crates/tpm2/src/crypto/hash.rs
+++ b/crates/tpm2/src/crypto/hash.rs
@@ -1,0 +1,203 @@
+use crate::{
+    Alg, TpmiAlgHash, TpmtHa,
+    crypto::{Base, Finalize, Update},
+};
+
+/// Cryptographic hashing interfaces for TPM implementations and clients.
+///
+/// Supported algorithms define a concrete `*Ctx` type and implement:
+///   - A method creating/initializing the context (e.g. [`Hash::sha256()`])
+///   - [`Update<Error>`] for `*Ctx`
+///   - [`Finalize<N, Error>`] for `*Ctx` (where `N` is the digest length)
+///
+/// Unimplemented algorithms bind their context type to [`!`] and
+/// omit the method, keeping the default body of
+/// [`Err(self.unimplemented(...))`](Base::unimplemented).
+///
+/// # Example
+///
+/// ```
+/// # use tpm2::{crypto::{Base, Hash, Finalize, Update}, Alg};
+/// # struct MyBackend;
+/// # struct MyError;
+/// # impl Base for MyBackend {
+/// #   type Error = MyError;
+/// #   fn unimplemented(&self, _: Alg) -> MyError { MyError }
+/// # }
+/// # struct MySha256;
+/// impl Hash for MyBackend {
+///     type Sha256Ctx = MySha256;
+///     fn sha256(&self) -> Result<MySha256, MyError> { todo!() }
+///
+///     // Unsupported algorithms use `!` and default methods:
+///     type Sha1Ctx = !;
+///     type Sha384Ctx = !;
+///     type Sha512Ctx = !;
+///     type Sm3_256Ctx = !;
+///     type Sha3_256Ctx = !;
+///     type Sha3_384Ctx = !;
+///     type Sha3_512Ctx = !;
+/// }
+///
+/// impl Update<MyError> for MySha256 {
+///     fn update(&mut self, data: &[u8]) -> Result<(), MyError> { todo!() }
+/// }
+/// impl Finalize<32, MyError> for MySha256 {
+///     fn finalize(self, out: &mut [u8; 32]) -> Result<(), MyError> { todo!() }
+/// }
+/// ```
+pub trait Hash: Base {
+    type Sha1Ctx: Update<Self::Error> + Finalize<20, Self::Error>;
+    fn sha1(&self) -> Result<Self::Sha1Ctx, Self::Error> {
+        Err(self.unimplemented(Alg::SHA1))
+    }
+
+    type Sha256Ctx: Update<Self::Error> + Finalize<32, Self::Error>;
+    fn sha256(&self) -> Result<Self::Sha256Ctx, Self::Error> {
+        Err(self.unimplemented(Alg::SHA256))
+    }
+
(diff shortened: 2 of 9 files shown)


crates/tpm2/src/crypto/hash.rs
```diff
@@ -0,0 +1,203 @@
+use crate::{
+    Alg, TpmiAlgHash, TpmtHa,
+    crypto::{Base, Finalize, Update},
+};
+
+/// Cryptographic hashing interfaces for TPM implementations and clients.
```
nnooney, line 6: This doc comment is great for implementors; we probably also need some lines for the users of the hash.

Not sure the best place where this belongs.

crates/tpm2/src/crypto/mod.rs
```diff
@@ -0,0 +1,30 @@
+//! Cryptography Interfaces for TPM implementations and clients
+use crate::Alg;
+
+mod hash;
+mod mac;
+pub use {hash::*, mac::*};
+
+pub trait Base {
```
nnooney, line 8: let's document these generic traits and their intended usage.

nnooney, line 8: Is Base the right name for this trait?

This seems related to handling errors from Crypto functions; do we want to define our own CryptoError trait?

crates/tpm2/src/crypto/mod.rs
```diff
@@ -0,0 +1,30 @@
...
+    type Error;
+
+    fn unimplemented(&self, alg: Alg) -> Self::Error;
+}
+
+pub trait Update<Error> {
+    fn update(&mut self, data: &[u8]) -> Result<(), Error>;
+}
+pub trait Finalize<const N: usize, Error> {
+    fn finalize(self, out: &mut [u8; N]) -> Result<(), Error>;
```
nnooney, line 18: It took me a bit of deep inspection to map the various finalize signatures, and I think it's worth mentioning:

"TPM APIs wrap this function to return TPM structs, for example Hash::Finalize wraps this trait function to return a TpmtHa".

crates/tpm2/src/crypto/hash.rs
```diff
@@ -0,0 +1,203 @@
...
+    }
+
+    type Sha3_512Ctx: Update<Self::Error> + Finalize<64, Self::Error>;
+    fn sha3_512(&self) -> Result<Self::Sha3_512Ctx, Self::Error> {
+        Err(self.unimplemented(Alg::SHA3_512))
+    }
+}
+
+/// Dynamic hash context wrapping algorithm-specific streams.
+pub enum HashCtx<H: Hash> {
```
nnooney, line 92: I think this might be more readable if we don't have 'Ctx' repeated in the trait types above and here.

In particular the new and finalize functions below handle wrapping trait inner ctx types and unwrapping; I think they may be clearer to understand if we call this `HashWrapper` or similar

crates/tpm2/src/crypto/hash.rs
```diff
@@ -0,0 +1,203 @@
...
+            HashCtx::Sha3_384(ctx) => ctx.update(data),
+            #[cfg(feature = "sha3_512")]
+            HashCtx::Sha3_512(ctx) => ctx.update(data),
+        }
+    }
+
+    /// Finalizes the digest into `out` and returns the tagged [`TpmtHa`].
+    pub fn finalize<'a>(
+        self,
+        out: &'a mut [u8; TpmtHa::MAX_DIGEST_SIZE],
```
nnooney, line 159: I don't have a better idea, but wondering if requiring out to have TpmtHa::MAX_DIGEST_SIZE would in practice cause the implementation to pass a larger buffer than necessary here; e.g. I can't pass a SHA256 sized buffer here if the MAX_DIGEST_SIZE also includes SHA384.

This might not be necessary in practice, since we always handle digests generically.

crates/tpm2/src/crypto/hash.rs
```diff
@@ -0,0 +1,203 @@
...
+    ) -> Result<TpmtHa<'a>, H::Error> {
+        /// Helper function for handling finalizing into the output buffer.
```
nnooney, line 161: nit: Helper function to bridge between a digest-specific sized buffer needed for finalize and the generic TpmtHa max buffer size.

I think the interesting part of this function is the treatment of buffer sizes more so than the wrapping around the generic finalize trait.

crates/tpm2/Cargo.toml
```diff
@@ -16,6 +16,16 @@ categories.workspace = true
 # Implement conversions/traits which require std
 std = []
 
+# Only enable SHA2 by default
+default = ["sha256", "sha384", "sha512"]
```
nnooney, line 20: we're probably going to need more default alg features later; let's update this comment to specify the default algorithms enabled by the crate.

```
# Default algorithms enabled. You probably don't want the default.
```

josephlr:
From @nnooney's comment in https://github.com/tpm-rs/tpm-rs/pull/262#discussion_r4046450662, we want to make sure `MAX_DIGEST_BYTES` gets updated to account for the enabled hash features.

crates/tpm2/src/crypto/hash.rs
josephlr: Added sections to differentiate between users and implementers of `Hash`.

crates/tpm2/src/crypto/mod.rs
josephlr: We just removed `Base` and used a common error type.

crates/tpm2/src/crypto/mod.rs
josephlr: I added:
```
/// Other traits and functions wrap this to return TPM structs. For example,
/// [`HashStream::finalize`] wraps this to return a [TpmtHa](crate::TpmtHa).
```

crates/tpm2/src/crypto/hash.rs
josephlr: Ya, it's not perfect, but because we handle things generically, it doesn't end up costing us any extra stack space.

crates/tpm2/Cargo.toml
josephlr: Changed defaults to just enable them all.

crates/tpm2/src/crypto/hash.rs
josephlr: Changed to HashStream, while all the other ones are now `*Context`

crates/tpm2/src/crypto/mod.rs
josephlr: Removed `Base`.

PR #262, opened 2026-09-09
Update `Tpm2b` and `Tpml` types to be generic and take buffers by reference

josephlr:
Refactors `TPM2B_*` byte buffers and `TPML_*` lists to replace macro-generated boilerplate with generic zero-copy/bounded containers (`Tpm2b<T>`, `Tpm2bSized<'a, Tag>`, and `Tpml<T, CAP>`).

### 1. `TPM2B_*` Refactoring (`Tpm2b<T>` and `Tpm2bSized<'a, Tag>`)
- **Structured vs. Byte Buffers**:
  - `Tpm2b<T>` wraps structured payloads (`Tpm2bAttest`, `Tpm2bCreationData`, ...).
  - `Option<Tpm2bSensitive<'a>>` implements `Marshal`/`Unmarshal` (`None` <-> `0u16`) to support public-only keys in `TPM2_LoadExternal` while keeping `Tpm2bSensitive` non-nullable elsewhere.
  - `Tpm2bSized<'a, Tag>` wraps raw byte slices bounded by a `limits::Tag`.
  - Both containers implement `Unmarshal` generically and share a generic `marshal_helper`, using a 1-line `impl_marshal!` macro per type alias for `Marshal` (until generic const expressions in associated types stabilize).
- **Nominal Capacity Tags (`limits::Tag`)**:
  - Parameterizes `Tpm2bSized<'a, Tag>` by zero-sized marker types implementing `limits::Tag { const CAP: usize; }` rather than `const CAP: usize`. This gives buffers with identical capacities (e.g., `Tpm2bEvent` and `Tpm2bMaxBuffer`) distinct nominal types.
  - Removes intermediate/unused wrapper types (`TpmsIdObject`, `TpmsContextData`, `Tpm2bContextSensitive`, `Tpm2bTemplate`) in favor of direct `Tpm2bSized` aliases (`Tpm2bIdObject`, `Tpm2bContextData`).
- **Centralized Buffer Limits (`structures::limits`)**:
  - **5 Vendor-Defined Constants**: `MAX_2B_BUFFER_SIZE`, `MAX_NV_BUFFER_SIZE`, `MAX_CAP_BUFFER`, `MAX_CONTEXT_SIZE` (`max(&[Private::CAP + TpmtPublic::MAX_SIZE + Tpm2bName::MAX_SIZE * 2, 4000])`), and `PRIVATE_VENDOR_SPECIFIC_BYTES` (`TpmiRsaKeyBits::MAX_PRIV_KEY_BYTES * 5`).
  - **Spec & Algorithm Limits**: All remaining `limits::Tag` capacities are either fixed by the specification (`Timeout`, `Label`, `SensitiveData`, `Event`), alias vendor constants (`MaxBuffer`, `MaxNvBuffer`), or derive automatically from enabled algorithms and structures (`Digest`, `Data`, `Name`, `Iv`, `SymKey`, `PublicKeyRsa`, `PrivateKeyRsa`, `EccParameter`, `EncryptedSecret`, `Private`, `IdObject`, `Context`).

### 2. Generic Fixed-Capacity Lists (`Tpml<T, CAP>`)
- Replaces macro-generated `TPML_*` list definitions with a generic `Tpml<T, const CAP: usize>` container.
- Uses `TpmlElement` (`const DEFAULT: Self`) so `Tpml::new(&[T]) -> Option<Self>` can initialize `[T::DEFAULT; CAP]` in `const` contexts without requiring `T: Default`.
- Derives list capacities directly from `MAX_CAP_DATA` (`limits::MAX_CAP_BUFFER - TpmCap::MAX_SIZE - u32::MAX_SIZE`) and `TpmiAlgHash::HASH_COUNT`.

### 3. Panic-Free Compiler-Visible Slice Bounds
Both `Tpm2bSized::as_slice` and `Tpml::as_slice` use `split_at_checked(Self::CAP)`:
```rust
if let Some((head, _)) = slice.split_at_checked(Self::CAP) {
    head
} else {
    slice
}
```
This proves `as_slice().len() <= CAP` to LLVM across all control-flow paths without `unsafe` code or panic branches in `#[no_std]` builds.


wenyuzhao (approved):

crates/tpm2/src/structures/tpm2b.rs
```diff
@@ -1,141 +1,189 @@
...
+impl<const N: usize, T: Marshal<MaxBuffer = [u8; N]>> Tpm2b<T> {
+    /// Helper constant to access the capacity of a `Tpm2b` type. For example,
+    /// [`Tpm2bPublic::CAP`] is equal to [`TpmtPublic::MAX_SIZE`].
+    pub const CAP: usize = N;
 
-/// Provides conversion to/from a struct type for TPM2B types that don't hold a bytes buffer.
-pub trait Tpm2bStruct: Tpm2bSimple {
-    type StructType: Marshal + for<'a> Unmarshal<'a>;
+    const fn max_size() -> usize {
+        u16::MAX_SIZE + N
```
josephlr: Just make this `2+N`

crates/tpm2/src/structures/tpm2b.rs
```diff
@@ -1,141 +1,189 @@
...
+    }
 
-    /// Marshals the value into the 2b holder.
-    fn from_struct(val: &Self::StructType) -> Result<Self, UnmarshalError>
-    where
-        Self: Sized;
+    fn marshal_helper<const M: usize>(&self, buf: &mut [u8; M]) -> usize {
+        // This check ensures these unwrap()/slicing ops don't panic.
+        const { assert!(M == Self::max_size()) };
+        let (head, rest) = buf.split_first_chunk_mut().unwrap();
```
josephlr:
```rust
buf.split_first_chunk_mut::<2>().unwrap();
```

crates/tpm2/src/structures/tpm2b.rs
```diff
@@ -1,141 +1,189 @@
...
+        let dst: &mut [u8; N] = (&mut rest[..N]).try_into().unwrap();
```
josephlr:
```rust
let dst: &mut [u8; N] = rest.try_into().unwrap();
```

crates/tpm2/src/structures/tpm2b.rs
```diff
@@ -1,141 +1,189 @@
...
+    fn unmarshal(src: &mut &'a [u8]) -> Result<Self, UnmarshalError> {
+        let len: usize = u16::unmarshal(src)?.into();
+        let mut buf: &'a [u8];
+        (buf, *src) = src.split_at_checked(len).ok_or(UnmarshalError)?;
+
+        let t = T::unmarshal(&mut buf)?;
+        match buf.len() {
+            0 => Ok(Self(t)),
+            _ => Err(UnmarshalError),
+        }
```
josephlr, line 43: Consistancy: should we be modifying `src` or not on failure?

crates/tpm2/src/structures/tpm2b.rs
```diff
@@ -1836,77 +450,108 @@ impl<'a> Unmarshal<'a> for Tpm2bContextData {
...
+    impl Tpm2bTag for Private {
+        const CAP: usize = Tpm2bDigest::MAX_SIZE * 2 + Tpm2bSensitive::MAX_SIZE;
+    }
+    /// Sized to hold an encrypted credential object (`sizeof(TPMS_ID_OBJECT)`).
+    pub struct IdObject;
+    impl Tpm2bTag for IdObject {
+        const CAP: usize = TpmsIdObject::MAX_SIZE;
+    }
+
+    /// Implementation-defined maximum size (`MAX_SENSITIVE_DATA`) for sealed data blobs.
```
josephlr: Rename constant to match reference impl / spec.

crates/tpm2/src/structures/tpm2b.rs
```diff
@@ -1836,77 +450,108 @@ impl<'a> Unmarshal<'a> for Tpm2bContextData {
...
-        x.creation_data[..sz].copy_from_slice(&max_buf[..sz]);
-        Ok(x)
-    }
-    fn to_struct(&self) -> Result<Self::StructType, UnmarshalError> {
-        tpm2b_to_struct(self)
+pub type Tpm2bCreationData<'a> = Tpm2b<TpmsCreationData<'a>>;
+impl_marshal!(Tpm2bCreationData<'_>);
+
+/// Marker tag types specifying the maximum byte capacity for [`Tpm2bSized`] buffers.
+pub mod tags {
```
josephlr: We should move this to its own file.

crates/tpm2/src/structures/tpm2b.rs
```diff
@@ -1836,77 +450,108 @@ impl<'a> Unmarshal<'a> for Tpm2bContextData {
...
+    /// Implementation-defined maximum bulk transfer size (`MAX_DIGEST_BUFFER`).
+    pub struct MaxBuffer;
+    impl Tpm2bTag for MaxBuffer {
+        const CAP: usize = 1024;
+    }
+    /// Implementation-defined maximum NV index read/write transfer size (`MAX_NV_BUFFER_SIZE`).
+    pub struct MaxNvBuffer;
+    impl Tpm2bTag for MaxNvBuffer {
+        const CAP: usize = 2048;
+    }
```
josephlr: What default value do we want here? 1024? With features to enable 2048/4096 sized NV Buffers?

crates/tpm2/src/structures/tpm2b.rs
```diff
@@ -1836,77 +450,108 @@ impl<'a> Unmarshal<'a> for Tpm2bContextData {
...
+    /// Implementation-defined maximum encrypted context size (`MAX_CONTEXT_SIZE`).
+    pub struct ContextSensitive;
+    impl Tpm2bTag for ContextSensitive {
+        const CAP: usize = 5120;
     }
```
josephlr: We should explicitly call out that this is "the gross one" which is set not based on the _features_ your TPM supports, but its internal implementation details.

crates/tpm2/src/structures/tpm2b.rs
josephlr: Resolved to just not require unmodified `src` on failure.

crates/tpm2/src/structures/tpm2b.rs
josephlr: Done by reworking `limits.rs`

crates/tpm2/src/structures/tpm2b.rs
josephlr: Defaulting to 1024, PC-client only requires 512.

josephlr:
@phil-levis-google @nnooney comments resolved and bugs fixed. I also cleaned up the tags/limits to now have all of that logic just live in `limits.rs`. PTAL.

phil-levis-google (approved):
Huzzah, no use after free and no unnecessary copying!

crates/tpm2-client/src/sessions/password.rs
```diff
@@ -17,41 +15,41 @@ use tpm2::{
...
     /// assert_eq!(
     ///     PasswordSession::new(&bad_password).err().unwrap(),
     ///     TpmRc::SIZE.to_rc()
     /// );
     /// ```
-    pub fn new<T: AsRef<[u8]> + ?Sized>(password: &T) -> Result<Self, TpmRc> {
+    pub fn new<T: AsRef<[u8]> + ?Sized>(password: &'a T) -> Result<Self, TpmRc> {
         Ok(PasswordSession {
-            auth: Tpm2bAuth::from_bytes(password.as_ref())?,
+            auth: Tpm2bAuth::new(password.as_ref()).ok_or(TpmRc::SIZE.to_rc())?,
```
nnooney, line 42: unfortunate that we need both `to_rc()` and the `?` operator here.

crates/tpm2/src/structures/tpmi.rs
```diff
@@ -218,6 +224,12 @@ pub enum TpmiAlgHash {
 }
 
 impl TpmiAlgHash {
+    /// The maximum digest size (in bytes) across all supported TPM2 hash algorithms.
+    pub const MAX_DIGEST_BYTES: usize = 64;
```
nnooney, line 228: Is this right? Shouldn't this be calculated based on the combination of features enabled?

crates/tpm2/src/structures/tpmt.rs
```diff
@@ -665,6 +663,10 @@ pub enum TpmtPublicParms {
 }
 
 impl TpmtPublicParms {
+    pub const MAX_SHARED_SECRET_BYTES: usize = TpmEccCurve::MAX_ECC_KEY_BYTES;
```
nnooney, line 666: The automatic constant forwarding here is very nice

nnooney (approved):

crates/tpm2-client/src/sessions/password.rs
josephlr: Changed to
```rust
pub fn new(password: &'a impl AsRef<[u8]> + ?Sized) -> Option<Self>
```
as we probably want to have the caller of such stuff handle any error information.

crates/tpm2/src/structures/tpmi.rs
josephlr: We'll address this as part of https://github.com/tpm-rs/tpm-rs/pull/261#issuecomment-5765284921

josephlr:
Added an additional doc comment as @nnooney suggested in #263:

Screenshot 2026-09-21 at 1 49 08 PM

